PR #890, opened 2026-05-18
fix(mobile): show correct day for daily menus on Android

malinmalliyawadu:
## What

Fixes daily menus rendering the **wrong day on Android** — e.g. Monday 18th's menu displayed as "Sunday 17th".

## Why

A menu's `serviceDate` is a *date-only* value (a calendar day), but it was being treated as a timestamp:

- The feed API (`web/src/app/api/mobile/feed/route.ts`) sends it as midnight-UTC ISO (`2026-05-18T00:00:00.000Z`).
- The mobile app ran it through `formatNZT()`, which does an `Intl`-based timezone conversion.

Converting midnight UTC rolls the day backward for any device behind UTC. On **Android specifically**, Hermes ships no IANA timezone data, so `@date-fns/tz`'s `tz("Pacific/Auckland")` silently falls back to *device-local* time — the shift happens regardless of intent. iOS and the web server have full ICU, so they rendered correctly and masked the bug.

## How

- Added `formatNZDateOnly()` to `mobile/lib/dates.ts` — reads the `YYYY-MM-DD` calendar fields straight off the value and formats them with **no timezone math**, so the result is identical on every OS and timezone. Accepts both the midnight-UTC ISO shape and a plain date string.
- Repointed all 5 `serviceDate` usages in `mobile/app/(tabs)/index.tsx` (feed-card title, expanded-sheet title, "past menu" check, and the "see shifts" CTA gate + date key) from `formatNZT(new Date(...))` to `formatNZDateOnly(...)`.
- `formatNZT` is unchanged and still used for genuine timestamps (shift times, "today").

Mobile-only fix — **no API, schema, or migration change**.

## Verification

`formatNZDateOnly` returns "Monday 18 May" / "2026-05-18" across `America/Los_Angeles` (the previously-failing behind-UTC case), `UTC`, and `Pacific/Auckland`, for both the ISO and plain-date input shapes.

## Notes for reviewers

- **Broader issue (out of scope):** the same Hermes-no-timezone-data limitation means *every* `formatNZT` call on Android renders in device-local time, not NZ. For real shift timestamps this is only visibly wrong if the device isn't on NZ time (fine for most local volunteers, wrong when travelling / on a misconfigured emulator). A platform-wide fix would add the `@formatjs/intl-datetimeformat` polyfill + NZ timezone data. This PR is scoped to the reported menu bug.
- Optionally the feed API could send date-only (`menu.date.toISOString().split("T")[0]`) to match the existing public `/api/menus` convention. The helper already handles both shapes, so this PR keeps the API untouched for minimal risk.

### DIFF
--- a/.github/workflows/mobile-ci.yml
+++ b/.github/workflows/mobile-ci.yml
@@ -1,0 +1,89 @@
+name: Mobile CI
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    branches: [main]
+    paths: ["mobile/**", ".github/workflows/mobile-ci.yml"]
+  push:
+    branches: [main]
+    paths: ["mobile/**", ".github/workflows/mobile-ci.yml"]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  TZ: Pacific/Auckland
+
+jobs:
+  mobile-test:
+    name: Mobile unit tests
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        working-directory: mobile
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          # Mobile has no own .nvmrc; reuse the repo's pinned Node version.
+          node-version-file: web/.nvmrc
+          cache: "npm"
+          cache-dependency-path: mobile/package-lock.json
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Unit tests
+        run: npm run test:run
+
+  # Re-emits the repo-wide "Tests passed" required check for mobile-only PRs,
+  # mirroring test.yml (web) and docs-check.yml (docs).
+  tests-passed:
+    name: Tests passed
+    runs-on: ubuntu-latest
+    needs: mobile-test
+    if: always()
+
+    steps:
+      - name: Check mobile tests
+        run: |
+          if [[ "${{ needs.mobile-test.result }}" == "success" ]]; then
+            echo "✅ Mobile unit tests passed"
+            exit 0
+          else
+            echo "❌ Mobile unit tests failed"
+            exit 1
+          fi
+
+  # Re-emits the repo-wide "Ready to Merge" required check for mobile-only PRs,
+  # mirroring ci.yml (web) and docs-check.yml (docs).
+  merge-check:
+    name: Ready to Merge
+    runs-on: ubuntu-latest
+    needs: mobile-test
+    if: always()
+
+    steps:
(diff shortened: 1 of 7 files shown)


malinmalliyawadu:
Thanks for the review. Addressed in a1408fb5:

- **Stricter validation** ✅ — `formatNZDateOnly` now rejects out-of-range parts (`Number.isInteger` + month 1–12 / day 1–31), so `"2026-13-45"` returns `""` instead of Date silently rolling it into the wrong day. (Note: `"invalid"`/`""` were already handled — `.map(Number)` yields `NaN`, which the truthiness check caught.)

Deliberately **not** changed, with rationale:

- **`string | Date` overload** — skipped intentionally. A `Date` input is exactly the ambiguous "instant vs. calendar day" case this helper exists to eliminate; accepting one would reintroduce the timezone footgun (`date.toISOString()` depends on how the Date was constructed). Keeping the input a string forces callers to pass the raw date-only value from the API.
- **Unit tests** — mobile has no test runner (no `test` script, no vitest/jest config; per CLAUDE.md unit tests are `web/`-only). A spec file would be dead code in CI. Behaviour was instead verified by running the helper across `America/Los_Angeles` / `UTC` / `Pacific/Auckland` for both input shapes plus the invalid cases. (FYI the suggested case `formatNZDateOnly("2026-05-18", ...)` → `"Sunday 18 May"` is off — 2026-05-18 is a Monday, which is what the helper correctly returns.)
- **Dev-only Android platform warning** — skipped as runtime noise; the broader `formatNZT`-on-Hermes limitation is already documented in the JSDoc and called out as a scoped follow-up in the PR description.

malinmalliyawadu:
Update: reversed my earlier "no tests" call and added a Vitest setup for mobile (commit `1ad8d216`).

- Added `vitest` (devDep), `mobile/vitest.config.ts` (node env, matching the web project's convention), and `test` / `test:run` scripts. Scoped to pure, framework-free modules — anything importing React Native still belongs in an Expo/RN runner.
- `mobile/lib/dates.test.ts` — 6 tests, all passing:
  - both input shapes (plain `YYYY-MM-DD` and midnight-UTC ISO)
  - year / month / leap-year boundaries
  - malformed + out-of-range input (`""`, `"invalid"`, `"2026-13-45"`, month/day 0, day 32)
  - **timezone-independence regression** — asserts `2026-05-18T00:00:00.000Z` still renders "Monday 18 May" under `America/Los_Angeles` (the original failing case), `UTC`, and `Pacific/Kiritimati` (UTC+14)
  - a `formatNZT` smoke test for instant → NZ conversion

Note: mobile isn't wired into CI (per CLAUDE.md, `ci.yml` is scoped to `web/**`), so these run locally via `npm run test:run` in `mobile/`. Wiring a mobile CI job is a reasonable follow-up but out of scope here.

PR #898, opened 2026-05-20
feat(notifications): push notifications for feed likes and comments

malinmalliyawadu:
# Pull Request

## Description
Sends push notifications when someone likes or comments on a user's feed post (achievement, friend signup, announcement), and when someone replies on a feed item the user has previously commented on. All push infrastructure already existed — `createNotification()` automatically fans to SSE + Expo — so this PR just adds three `NotificationType` enum values and wires them into the like and comment routes.

### What's notified

| Event | Recipient | Type | Coalescing |
|---|---|---|---|
| Like on your post | post owner | `FEED_ITEM_LIKED` | While unread, updates in place to "X and N others liked your post" and re-broadcasts SSE without a fresh push (avoids spam on a popular post). Read → next like fires a new push. |
| Comment on your post | post owner | `FEED_ITEM_COMMENTED` | One per comment. 80-char snippet in body. |
| Comment on a post you commented on | other thread participants | `FEED_ITEM_COMMENT_REPLY` | One per comment, batched via `createNotificationsForUsers` (Expo chunks of 100 with per-user badge counts). Excludes the actor and the post owner (who already gets `FEED_ITEM_COMMENTED`). |

### Owner resolution by feed item prefix
- `achievement-*` → `UserAchievement.userId` ("your achievement")
- `friend-signup-*` → `Signup.userId` ("your shift signup")
- `announcement-*` → `Announcement.createdBy` ("your announcement")
- `new-shift-*`, `shift-recap-*`, `daily-menu-*` → no owner, skipped

Self-actions (actor == owner) are filtered out at the route layer.

## Type of Change

## Version Bump Required
- [x] `version:minor` — new feature

## Testing
- [x] Unit tests pass (`vitest run src/lib/feed/feed-item-owner.test.ts` — 6/6)
- [x] Typecheck passes (`npm run typecheck`)
- [x] Lint clean on touched files (`npm run lint`)
- [ ] Manual end-to-end on simulator + real device (requires dev env)
- [x] New tests added — `web/src/lib/feed/feed-item-owner.test.ts` covers each prefix, missing record, system-generated items, and unknown prefixes

### Reviewer test plan
1. Apply migration: `cd web && npm run prisma:migrate`.
2. Log in as User A on a real device with Expo push registered (check `PushToken` row).
3. As User B, like User A's achievement → User A receives one push.
4. As User B, unlike + relike, and User C also likes → A's existing unread notification updates to "and 2 others", **no second push fires**.
5. A marks the notification read; a subsequent like fires a fresh push.
6. As User B, comment → A receives a `FEED_ITEM_COMMENTED` push with the snippet.
7. As User C, comment on the same item → both A (`FEED_ITEM_COMMENTED`) and B (`FEED_ITEM_COMMENT_REPLY`) receive pushes; C does not.
8. As User A, like own item → no notification, no push.

## Checklist
- [x] Code is commented only where the WHY isn't obvious (coalescing logic)
- [x] No new warnings introduced
- [x] New unit tests added for the feed-item owner resolver
- [x] All unit tests pass locally

## Additional Notes
- Migration adds three values to the `NotificationType` enum in one file using the established `ALTER TYPE ... ADD VALUE` pattern (matches existing migrations like `20250918070325_add_nova_signup_statuses`).
- No opt-out toggles in v1 — matches existing notification types, which don't expose per-user prefs either. Easy to add later via `User.notifyOnFeedLikes` etc.
- Mobile already deep-links `/dashboard?feedItemId=X` via `mobile/lib/notification-routing.ts:31`, so no routing changes needed.
- The mobile `NotificationType` union was previously out of sync with the server (missing `SHIFT_SHORTAGE`, `ANNOUNCEMENT`, `DIRECT_MESSAGE`) — left as-is to keep this PR focused; worth a follow-up.


malinmalliyawadu:
Thanks for the review! Triaged the suggestions:

**Addressed in 3d58a1fd:**
- Extracted `COMMENT_SNIPPET_MAX_LENGTH = 80` + `snippetOf()` helper so the truncation logic isn't duplicated across `notifyFeedItemCommented` and `notifyFeedItemCommentReply`.

**Declined (with reasoning):**
- *Parallelize owner + actor lookup in the like route:* the current serial path early-returns when there's no owner or the actor is liking their own item, so it makes 1 query in the cold path vs 2 in the suggested version. The notification is fire-and-forget, so the ~5ms saved in the hot path doesn't matter. Net cost.
- *`COUNT(*)` on every like:* `FeedLike` has `@@index([feedItemId])`, so this is a cheap index-only scan — not a hot-path concern.
- *Race condition in like coalescing:* real, but the impact is bounded — two simultaneous likers each see `existing=null` and create two notifications instead of one coalesced. The fix is a partial unique index on `(userId, type, relatedId) WHERE isRead=false`, which adds schema complexity for a tiny UX delta (one extra push in the rare race window). Not worth it.
- *Extract `notifyFeedOwnerOfLike` to a utilities module:* premature with a single caller. The like and comment routes use different notification flows (coalescing vs fan-out), so a shared helper would have to handle both shapes.
- *`FEED_ITEM_ID_MAX_LENGTH = 200` constant:* the literal only appears once per route — extracting it doesn't DRY anything up.

**Out of scope (already noted in PR description):**
- Mobile `NotificationType` union sync with server enum — pre-existing gap, worth a follow-up issue.

Issue #902, opened 2026-05-22
[Web] DOMException HierarchyRequestError: new child element contains the parent on /shifts/details


malinmalliyawadu:
## Error

`DOMException: HierarchyRequestError: Failed to execute 'insertBefore' on 'Node': The new child element contains the parent.`

## Top stack frame

```
$RV  /shifts/details:1:28710  (in-app, minified — source map resolution failed: bad JSON)
```

<details>
<summary>Full stack trace</summary>

```
DOMException: HierarchyRequestError: Failed to execute 'insertBefore' on 'Node': The new child element contains the parent.

  at $RV (https://volunteers.everybodyeats.nz/shifts/details:1:28710)
```

_Source map resolution failed ("Invalid source map: bad json: expected value at line 1 column 1"), so only the minified frame is available. The function `$RV` is a React internal (likely `commitMutationEffects` or similar reconciler code)._

</details>

## Impact (last 24h)

| Metric | Value |
|--------|-------|
| Occurrences | 13 |
| Affected users | 3 |
| Sessions | 3 |
| First seen | 2026-05-21 09:14:48 UTC |
| Last seen | 2026-05-22 18:27:49 NZST |

## Reproduction context

- **URL**: `https://volunteers.everybodyeats.nz/shifts/details?date=2026-06-11&location=Onehunga`
- **Browser**: Chrome 148 on Windows 10
- **Library**: `web` (posthog-js)

## Notes

The error is a circular DOM reference — a node is being inserted as a child of itself (or one of its own descendants). This is typically caused by a React reconciliation bug, often triggered by:

- Incorrect `key` props on list items causing React to mis-match old and new trees
- A component being conditionally rendered in a way that confuses the reconciler
- Server/client HTML mismatch (hydration) on the `/shifts/details` page

The source map for `/shifts/details` is returning invalid JSON, which is also worth fixing as it blocks debugging.

## PostHog

[View issue in PostHog](https://us.posthog.com/project/211852/error_tracking/019e49d1-2b77-70f2-a886-3d9e1e863ce5)

PR #905, opened 2026-05-22
fix(shifts): prevent HierarchyRequestError on rapid date navigation

malinmalliyawadu:
## Summary

- Adds a stable `key` to the `<Suspense>` boundary that wraps `ShiftDetailsContent` on `/shifts/details`, derived from the current `date` and `location` search params.
- Prevents `DOMException: HierarchyRequestError: Failed to execute 'insertBefore' on 'Node': The new child element contains the parent.` thrown during React commit when users navigate prev/next day rapidly.

## Why

[Issue #902](https://github.com/everybody-eats-nz/volunteer-portal/issues/902) — PostHog captured 13 occurrences from 3 users over 24h, all during rapid date navigation on `/shifts/details?location=Onehunga`. The crash happens inside React's `commitMutationEffects` (the internal `$RV` frame).

Without a key, the same `<Suspense>` instance is reused across navigations. Its subtree contains Radix `Dialog`/`Drawer` portals (inside every `ShiftSignupButton`/`CancelSignupButton`), and the surrounding page is wrapped in animated `motion.div` containers. When `searchParams` change rapidly, React tries to reconcile the old portal-using subtree into the new tree and, in some interleavings, ends up scheduling a DOM insert that would place a node inside one of its own descendants — the browser rejects this with `HierarchyRequestError`.

Forcing a clean unmount/remount via `key={\`\${dateParam}-\${selectedLocation ?? "all"}\`}` sidesteps the reconciliation conflict entirely. The admin-side `animated-shift-cards-wrapper.tsx` already uses the same `${dateString}-${selectedLocation}` keying pattern (via `AnimatePresence mode="wait"`) for the same scenario.

## Test plan

- [ ] Open `/shifts/details?date=YYYY-MM-DD&location=Onehunga` (or any location).
- [ ] Click the prev/next day buttons rapidly several times in a row.
- [ ] Confirm no `HierarchyRequestError` appears in the browser console.
- [ ] Confirm the shift list updates correctly for each date and the skeleton fallback appears briefly between navigations.
- [ ] Open a `ShiftSignupButton` or `CancelSignupButton` dialog on a shift card and confirm it still opens/closes normally.
- [ ] Repeat on mobile width to confirm the Drawer variant still works.

### DIFF
--- a/web/src/app/shifts/details/page.tsx
+++ b/web/src/app/shifts/details/page.tsx
@@ -143,8 +143,14 @@ export default async function ShiftDetailsPage({
         <ShiftsProfileCompletionBanner />
       </Suspense>
 
-      {/* Shift cards stream in when data is ready */}
+      {/* Shift cards stream in when data is ready.
+          Key forces a clean unmount/remount on navigation between dates or
+          locations — without it, rapid prev/next clicks can put React's
+          reconciler into a state where it tries to re-parent a portal-using
+          subtree (Radix Dialog/Drawer inside ShiftSignupButton) into one of
+          its own descendants, throwing HierarchyRequestError during commit. */}
       <Suspense
+        key={`${dateParam}-${selectedLocation ?? "all"}`}
         fallback={
           <div className="space-y-8">
             <div className="space-y-4">


malinmalliyawadu:
The failing **Playwright Tests (5)** shard is an unrelated flake in `admin-shift-edit-delete.spec.ts:331` ("should delete shift successfully when confirming"). This PR only touches `web/src/app/shifts/details/page.tsx` (volunteer-facing), not the admin shifts page or that test.

The failure is a strict-mode violation:

```
strict mode violation: getByTestId('shift-deleted-message') resolved to 2 elements:
    1) <div data-testid="shift-deleted-message" ...>Shift deleted successfully!</div>
    2) <div data-testid="shift-deleted-message" ...>Shift deleted successfully!</div>
```

Both matches have identical HTML — the page DOM transiently contains the alert twice during the post-delete redirect (`window.location.href = "/admin/shifts?deleted=1&..."`). The same test had a similar strict-mode flake fixed in #884 last week; this looks like a re-emergence under different timing (cacheComponents/Suspense streaming).

All 19 other shards pass. Happy to address as a separate follow-up if needed, but it's out of scope for the `/shifts/details` HierarchyRequestError fix.

PR #907, opened 2026-05-22
feat(shifts): shareable shift pages with dynamic OG images

malinmalliyawadu:
## Description

Improves shareability of individual shift pages so volunteers can post a link and have it render with a rich, branded preview on social platforms and messaging apps.

## What changed

- **Share button** ([web/src/components/share-shift-button.tsx](web/src/components/share-shift-button.tsx)) — Web Share API on mobile, copy-to-clipboard with a sonner toast on desktop. Shows a "Copied" check icon for ~2s after copy. Placed in a new "Spread the word" section beneath the calendar export buttons on the shift detail page.
- **Dynamic OG image** ([web/src/app/shifts/[id]/opengraph-image.tsx](web/src/app/shifts/[id]/opengraph-image.tsx)) — 1200×630 branded card rendered by `next/og` `ImageResponse`. Shows shift type heading (Fraunces), NZ-formatted day/date/time (Libre Franklin), location, and a status pill (`X spots left` / `Waitlist open` / `Shift complete`). Theme gradient blobs are colour-matched per shift type via `shift-themes.ts`. Fonts are fetched from Google Fonts at render time with a satori-compatible UA, with a graceful fallback if the fetch fails.
- **`generateMetadata`** on [web/src/app/shifts/[id]/page.tsx](web/src/app/shifts/[id]/page.tsx) — Per-shift title (e.g. `Dishwasher · Wellington · Friday 23 May`), description that includes the spots-remaining state, canonical URL via the existing `buildPageMetadata` helper. Past shifts are marked `noindex` so dead pages don't accumulate in search results.
- **Schema.org `Event` JSON-LD** emitted from the page using the existing `buildShiftEventSchema` helper — gives Google rich snippets for shift events.
- **Sitemap** ([web/src/app/sitemap.ts](web/src/app/sitemap.ts)) — Upcoming shift URLs (start ≥ now, capped at 5000) added with `lastModified = shift.start` and priority 0.5, so individual shifts become discoverable to crawlers.

## Type of Change

## Testing

- [x] `npm run typecheck` passes
- [x] `npm run lint` — 0 errors on changed files (only pre-existing repo warnings remain)
- [x] `next build` compiles all routes successfully (subsequent build-time data-collection failure is unrelated — pre-existing top-level Prisma call in `web/src/lib/locations.ts` that needs a real DB at build time)
- [ ] Manual preview testing — verify share button, OG preview, and sitemap entries in deployed preview

## How to test locally

1. `cd web && npm run dev`
2. Visit any shift page at `/shifts/<id>` and try the Share button (works both on mobile via native share sheet and on desktop via clipboard copy)
3. Visit `/shifts/<id>/opengraph-image` directly to preview the OG card
4. Paste a deployed preview URL into [opengraph.xyz](https://www.opengraph.xyz/) (or post in Slack / WhatsApp) to verify the OG image renders correctly

## Notes

- Share button and the "Spread the word" section are only rendered for upcoming shifts (matches existing calendar-export visibility), so we don't promote sharing of dead links.
- Description messaging uses warm te reo-flavoured tone consistent with the brand voice ("Kia ora — join the whānau", "Spread the word", etc.).
- No new dependencies — uses `next/og` (already part of Next.js) and existing `sonner`, `lucide-react`, and shadcn `Button`.

### DIFF
--- a/web/src/app/shifts/[id]/opengraph-image.tsx
+++ b/web/src/app/shifts/[id]/opengraph-image.tsx
@@ -3,7 +3,6 @@ import { prisma } from "@/lib/prisma";
 import { formatInNZT } from "@/lib/timezone";
 import { getShiftTheme } from "@/lib/shift-themes";
 
-export const runtime = "nodejs";
 export const alt = "Volunteer shift at Everybody Eats";
 export const size = { width: 1200, height: 630 };
 export const contentType = "image/png";
--- a/web/src/app/shifts/[id]/opengraph-image.tsx
+++ b/web/src/app/shifts/[id]/opengraph-image.tsx
@@ -29,9 +29,9 @@ async function loadGoogleFont(family: string, weight: number): Promise<ArrayBuff
       `https://fonts.googleapis.com/css2?family=${encodeURIComponent(family)}:wght@${weight}&display=swap`,
       {
         headers: {
-          // UA pinned so Google returns a satori-compatible font format (TTF/WOFF).
+          // Pre-woff2 UA so Google returns TTF, which is what satori (next/og) supports.
           "User-Agent":
-            "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36",
+            "Mozilla/5.0 (Windows NT 6.1; Trident/7.0; rv:11.0) like Gecko",
         },
       }
     ).then((r) => r.text());
@@ -334,7 +334,7 @@ export default async function ShiftOgImage({
             display: "flex",
             alignItems: "center",
             justifyContent: "space-between",
-            padding: "0 72px 56px",
+            padding: "32px 72px 56px",
           }}
         >
           <div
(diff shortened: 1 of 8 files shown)


malinmalliyawadu:
Thanks @claude for the review. Pushed [6689dc4](https://github.com/everybody-eats-nz/volunteer-portal/commit/6689dc42) to address it:

**Addressed**

- **Font fetch timeout & URL allowlist** — Extracted both copies of `loadGoogleFont` into [web/src/lib/og-fonts.ts](https://github.com/everybody-eats-nz/volunteer-portal/blob/claude/blissful-ardinghelli-773ed5/web/src/lib/og-fonts.ts). It now uses an `AbortController` with a 4s timeout and the regex is anchored to `https://fonts.gstatic.com/` so a malformed CSS response can't redirect us at an arbitrary host. Allowed families are typed as a string-literal union.
- **Sitemap query** — Switched `start >= now` to `end >= now` so Postgres uses the existing `@@index([end])` on `Shift` rather than a sequential scan. As a side effect, in-progress shifts stay shareable until they wrap up, which is the correct behaviour.
- **Sitemap caching** — Added `export const revalidate = 3600` so the sitemap is cached for an hour between crawler hits.

**Not addressing in this PR**

- *Theme color fallback defensiveness* — `THEME_PRESETS["from-gray-500 to-slate-500"]` is a hardcoded object literal one block above the lookup, so it can't be undefined at runtime. Adding a second fallback would be dead code.
- *Font caching in memory* — `next/og` ImageResponse output is already cached by Next.js per-URL, so the Google Fonts request only fires on cold/regenerated images. An additional process-level font cache would be premature.
- *Magic-number constants for OG dimensions* — Already defined as `size`/`SIZE` constants at the top of each route; further extraction is cosmetic.
- *Test coverage* — Worth doing as a follow-up but out of scope here. Will track separately.

PR #940, opened 2026-06-03
feat(admin): clickable engagement stat cards — drill into the volunteers

malinmalliyawadu:
## Description

Follow-up to #939. Every stat card on the **Volunteer Engagement** report is now **clickable** — clicking a card opens a modal listing the volunteers behind that number, so admins can see exactly *who* and follow up.

Clickable cards: **Total, Retention, New, Reactivated**, and the four status cards **Highly Active / Active / Inactive / Never**.

Each modal row shows the volunteer (with link to their profile), their last shift, shifts-in-period and total shifts. The Reactivated list additionally shows when they came back, when they were last seen, and the gap length. Every list is scoped to the same period / location / day-of-week filters as the cards, so the list and the KPI always agree.

## How it's built (one generic system, not eight one-offs)
- **`getEngagementSegmentVolunteers(segment, …)`** in `engagement.ts` handles all segments. It reuses the same classification as the existing on-page Volunteers table (so counts line up) and delegates the reactivated segment to `getReactivatedVolunteers()`. Capped at 500 rows with an accurate total.
- **One** ADMIN-gated route `/api/admin/analytics/engagement/segment` (validates the `segment` param).
- **One** reusable `EngagementSegmentDialog` (per-segment title, subtitle, avatar accent), with loading skeletons, empty and error states.
- All hero tiles and category cards are now accessible `<button>`s (keyboard-focusable, `aria-label`).

## Type of Change

## Version Bump Required
`version:minor`

## Testing
- [x] Unit tests pass (208); `npm run typecheck` and `npm run lint` clean (0 errors)
- [x] Manual verification against the dev DB (months=12): every segment endpoint returns 200, and each segment total matches its KPI exactly — Total 83, Highly Active 1, Active 82, Inactive 0, Never 0, New 83, Retention 0, Reactivated 0 (83 = 1 + 82). Opened the Active modal (82 rows with last-shift / in-period / total badges) and confirmed the empty-state for a zero segment (Never). Reactivated verified earlier via a reversible data scenario.

## Reviewer notes
- Segment counts reuse the existing table's classification, so for **All Locations** they match the KPIs exactly; for a location filter there can be tiny edge differences for volunteers associated with multiple locations (same caveat as the existing table).
- Uses the applied filters (not in-progress selections) so each list matches what's on screen.

### DIFF
--- a/web/src/app/admin/analytics/engagement/engagement-analytics-client.tsx
+++ b/web/src/app/admin/analytics/engagement/engagement-analytics-client.tsx
@@ -45,7 +45,8 @@ import {
 } from "@/components/ui/dialog";
 import { DayOfWeekFilter } from "@/components/day-of-week-filter";
 import { EngagementVolunteerTable } from "./engagement-volunteer-table";
-import { ReactivatedVolunteersDialog } from "./reactivated-volunteers-dialog";
+import { EngagementSegmentDialog } from "./engagement-segment-dialog";
+import type { EngagementSegment } from "@/lib/engagement";
 import type {
   EngagementSummaryData,
   EngagementVolunteersResult,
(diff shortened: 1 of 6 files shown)


malinmalliyawadu:
Thanks for the review! Applied suggestion 1 — the reactivation gap now uses `date-fns` `differenceInMonths` for accurate whole-month boundaries instead of dividing by an average 30.44 days (commit b3021ac).

On the other two: the generic "Failed to load volunteers" message is intentionally simple and the dialog already surfaces the server error detail when present, so leaving as-is. The composite `(Shift.end, Shift.location)` index isn't worth a migration here — the drill-down is an admin-only, ≤500-row query and `Shift` already has an `@@index([end])`; happy to add it later if these reports show up slow in practice.

PR #985, opened 2026-06-15
feat(admin): export service-night analytics as CSV

malinmalliyawadu:
## What

Adds an **Export CSV** button to the `/admin/analytics` filter toolbar (next to *Apply Filters*) that downloads the service-night data for the **currently applied filters**.

## Why

Makes it easy to pull the restaurant data out for ad-hoc analysis (e.g. uploading to a spreadsheet or an LLM's analysis tool) without hand-querying the DB. The on-screen table is paginated; the export gives the full filtered dataset in one file.

## How

- New `?format=csv` mode on `GET /api/admin/analytics/service-nights` returns **all** matching rows (ignores pagination), respecting the same filters (time period, location, day-of-week, custom date range) and sort.
- Richer columns than the table — includes the full koha split (cash / eftpos / quest-stripe) plus vege, takeaways, eftpos transactions and notes — ideal for analysis.
- RFC-4180 cell quoting + UTF-8 BOM so Excel/Google Sheets open it cleanly. Sensible filename, e.g. `ee-service-nights_all-time_Wellington.csv`.
- Button exports the **applied** filters (what's on screen), shows a loading state, and toasts the row count. Click *Apply Filters* first if you've changed a filter.

## Verification

Typecheck: 0 errors. Lint: 0 errors / baseline warnings. Verified end-to-end locally — button renders in the filter toolbar, exporting all-time Wellington produced 852 rows with the correct header/koha-split and a success toast, no console errors.

### DIFF
--- a/web/src/app/admin/analytics/restaurant-analytics-client.tsx
+++ b/web/src/app/admin/analytics/restaurant-analytics-client.tsx
@@ -40,7 +40,9 @@ import {
   Salad,
   Beef,
   CloudSun,
+  Download,
 } from "lucide-react";
+import { toast } from "sonner";
 import dynamic from "next/dynamic";
 import {
   Tooltip,
@@ -245,6 +247,7 @@ export function RestaurantAnalyticsClient({
   const [to, setTo] = useState(initialTo);
   const [trendView, setTrendView] = useState<"monthly" | "weekly">("monthly");
   const [kohaStack, setKohaStack] = useState<StackMode>("stacked");
+  const [exporting, setExporting] = useState(false);
   const chartThemeMode = (resolvedTheme === "dark" ? "dark" : "light") as
     | "dark"
     | "light";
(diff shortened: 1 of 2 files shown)


malinmalliyawadu:
Thanks for the review 🙏 Addressed the security finding and added tests; rationale on the rest below.

**Fixed (pushed in `3e2a815`):**
- **CSV formula injection** — extracted the CSV helpers into `src/lib/csv.ts` and `csvCell` now prefixes string cells that lead with `=`, `+`, `-`, `@` (or a control char) with a single quote, so admin free-text (`notes`/`protein`/`weather`) can't execute as a formula in Excel/Sheets. Numbers are emitted verbatim so negatives aren't mangled.
- **Test coverage** — added `src/lib/csv.test.ts` (9 cases) covering RFC-4180 quoting/escaping and the injection guard, including the guard-plus-quote combo.

**Considered, not changed (happy to if you'd prefer):**
- *Row-count edge case* — current `text.trim().split("\n").length - 1` already yields `0` for a headers-only response (BOM is stripped by `trim`), so behaviour is correct.
- *Empty result / 204* — returning a headers-only CSV is the expected UX for "no rows match"; a 204 would make the download silently do nothing.
- *Large-dataset streaming / row-count warning* — the full dataset is ~2.7k rows (low tens of KB), well within a single in-memory response; not worth the complexity yet.
- *Rate limiting* — endpoint is ADMIN-only behind session auth, so abuse surface is minimal; can add `src/lib/rate-limit.ts` if desired.
- *`console.error` / `handleExport` length / JSDoc* — matches existing patterns in the codebase; left as-is.
- *E2e test* — deferred; the security-critical logic is now unit-tested.

Typecheck: 0 errors · Lint: 0 errors · `csv.test.ts`: 9/9 passing.

Issue #11, opened 2025-08-13
Add e2e tests for Admin Shifts Management


malinmalliyawadu:
## Summary
Create comprehensive e2e tests for admin shift management pages (`/admin/shifts`, `/admin/shifts/new`, `/admin/shifts/[id]/edit`) to ensure admins can create and manage shifts effectively.

### Admin Shifts List (`/admin/shifts`)
- [ ] Page displays all shifts in admin view
- [ ] Shift filtering and search functionality
- [ ] Edit and delete actions for existing shifts
- [ ] Create new shift button functionality
- [ ] Shift status indicators and management
- [ ] Bulk actions if applicable
- [ ] Pagination for large shift lists

### Create New Shift (`/admin/shifts/new`)
- [ ] Shift creation form displays correctly
- [ ] All required fields validation
- [ ] Date and time picker functionality
- [ ] Location selection
- [ ] Shift type/category selection
- [ ] Volunteer capacity settings
- [ ] Description and requirements fields
- [ ] Save and cancel functionality
- [ ] Success/error handling

### Edit Shift (`/admin/shifts/[id]/edit`)
- [ ] Form pre-populates with existing shift data
- [ ] All fields can be modified
- [ ] Update functionality works
- [ ] Delete shift functionality
- [ ] Confirmation dialogs for destructive actions
- [ ] Impact on existing signups handling

## Acceptance Criteria
- [ ] Admin-only access enforced on all pages
- [ ] CRUD operations work correctly for shifts
- [ ] Form validation prevents invalid data
- [ ] Existing volunteer signups handled appropriately
- [ ] Tests use testid attributes consistently
- [ ] Error scenarios handled gracefully

## Priority
**Medium** - Critical admin functionality

## Implementation Notes
- Test with various shift data scenarios
- Verify admin role requirements
- Test form validation thoroughly
- Use testids for all interactive elements
- Test impact on volunteer signups when editing/deleting shifts

PR #1000, opened 2026-06-18
feat(mobile): in-app registration with scroll-to-read agreements & Apple/Google sign-up

malinmalliyawadu:
## Description

Builds a native in-app registration flow for the mobile app. Previously the login screen's "Join our whānau" link just opened the website; now volunteers can sign up entirely inside the app — by email or with Apple/Google — and the required agreements are shown and must be read before acceptance.

### What changed

**Native registration screen** (`mobile/app/(auth)/register.tsx`)
- Email sign-up matching the web `/api/auth/register` contract: name, email, password (live requirements mirroring the server rules), confirm-password, and the two required agreements.
- `AuthGate` swaps between login/register with local state (the unauthenticated tree has no navigator mounted), and the login footer now opens this screen instead of the website.

**Dedicated mobile register endpoint** (`/api/auth/mobile/register`)
- Mirrors the web registration side-effects (verification email, newsletter sync, `profileCompleted`) but skips Cloudflare Turnstile (the native app can't produce a token) and rate-limits by IP instead. Returns a session token directly so the app logs in immediately.

**Scroll-to-read agreements**
- Each agreement opens a full-screen reader that renders the canonical policy markdown — fetched from the **same `/content/volunteer-agreement.md` / `/content/health-safety-policy.md` the web serves**, so wording can't drift. The "I agree" button stays locked until the volunteer scrolls to the end. Tiny self-contained markdown renderer, no new dependency.

**Apple / Google sign-up + agreement gate**
- Added Apple/Google to the register screen (OAuth sign-in doubles as sign-up — the backend upserts on first auth).
- Because OAuth account creation can't capture the agreements, a new `agreementsAccepted` flag flows through every mobile auth response (`toMobileUser`), and `AuthGate` routes any authenticated account that hasn't accepted both through a dedicated agreement gate before entering the app. Acceptance is persisted via `POST /api/auth/mobile/agreements`. The gate shows exactly once and email sign-ups skip it (they accept during registration).

## Type of Change

## Version Bump Required
`version:minor`

## Testing
- [x] Web typecheck (`tsc --noEmit`) + ESLint pass
- [x] Mobile typecheck passes
- [x] Full iOS Metro bundle (`expo export`) succeeds
- [ ] Manual end-to-end testing on simulator (pending)

## Additional Notes
- The merge of `main` (mobile brand restyle) is included; the register hero image was repointed to a current asset after `main` removed the old one.
- Mobile sign-ups intentionally collect only the API-required fields and start `profileCompleted: false`; the remaining profile fields (DOB, emergency contact, etc.) are completed in-app, gated the same way the web gates shift signup.
- Heads-up unrelated to this PR: `mobile/package.json` pins `react-native@0.86.0` against the Expo-55 stack (codegen/gesture-handler expect ~0.83), so clean installs break a Metro bundle until `react-native` is realigned to 0.83.x. Not addressed here.


malinmalliyawadu:
Thanks for the thorough review! Here's how I've addressed it.

**CI fix (the actual failure):** The failing `lint` job was a unit-test failure, not an ESLint error — adding `agreementsAccepted` to `toMobileUser` changed its output shape and broke the equality assertion in `mobile-auth.test.ts`. Updated that assertion and added a case asserting the flag is only `true` when both agreements are accepted (`aa74b098`). Full suite green locally (264 passed).

**Already handled by existing design:**
- **#1 HTTPS policy fetch** — `API_URL` is a hardcoded `https://volunteers.everybodyeats.nz` constant (`mobile/lib/api.ts`), so policy fetches are already HTTPS-only; there's no http fallback to downgrade.
- **#2 CSRF on `/agreements`** — this is a Bearer-token API (`Authorization` header, no ambient cookie auth), so it isn't reachable by a CSRF cross-site POST. Left as-is.
- **#8 fetch-failure feedback** — tapping an agreement opens the reader, which already renders an error state + "Try again" when the markdown fails to load (`agreement-modal.tsx`).

**Pre-existing / intentional, out of scope here:**
- **#4 Google OAuth race** — the `googleInFlight` ref pattern was extracted verbatim from the existing login screen (shared via `useOAuthSignIn`), not introduced here. Worth a separate hardening pass since it'd touch both entry points.
- **#5 short-policy auto-unlock** — both policy docs are ~6KB and never fit on screen, so the scroll-to-end gate always engages in practice; the auto-unlock path only applies to hypothetical short content.
- **#6 `profileCompleted` OR-fallback** and **#11 newsletter default `true`** — both mirror existing `toMobileUser` / web-register behavior intentionally, for parity.

**Follow-ups (noted, not in this PR):** broader API-route tests for `/mobile/register` and `/mobile/agreements` (#9), and splitting the large `register.tsx` into form subcomponents (#11). Happy to do these in a follow-up if preferred.

PR #1001, opened 2026-06-18
fix(auth): verify passkey assertion in-process to fix prod login timeout

malinmalliyawadu:
## Description

Passkey login was failing in production with a connect timeout:

```
Passkey authentication error: [TypeError: fetch failed] {
  [cause]: Error [ConnectTimeoutError]: Connect Timeout Error
      (attempted address: volunteers.everybodyeats.nz:443, timeout: 10000ms)
}
```

**Root cause:** the NextAuth passkey provider verified assertions by making an HTTP `fetch()` to `${NEXTAUTH_URL}/api/passkey/authenticate/verify-authentication` — i.e. the server calling its own public hostname. On the new Coolify/Vultr (Sydney) infra there is no hairpin-NAT loopback (the proxy sits in front), so the server cannot route back to its own public URL and the request hangs until undici's 10s timeout. This worked on Vercel due to how it handled the loopback, but broke after the migration.

**Fix:** extract the full verification ceremony into a shared in-process function, `verifyPasskeyAuthentication()` in `web/src/lib/passkey-auth.ts`, and call it directly from:
- the NextAuth passkey `authorize()` callback in `auth-options.ts` (no more network hop), and
- the HTTP route `verify-authentication/route.ts`, now a thin wrapper that delegates to the same function.

No behavioural change to the verification logic itself — same challenge consumption, assertion verification, anti-cloning counter check, counter/`lastUsedAt` update, and archived-user reactivation.

## Type of Change

## Version Bump Required
`version:patch`

## Testing
- [x] `tsc --noEmit` passes on all changed files (one pre-existing, unrelated error remains in `restaurant-reports.tsx`)
- [ ] Live passkey login not exercisable locally — WebAuthn needs real authenticator hardware and the prod RP origin. Verified by type-checking + code review; confirm on the deploy.

## Additional Notes
- Scanned the rest of `src/` — this was the **only** server-side self-fetch to the app's own public URL, so no other feature carries the same latent bug.
- The mobile passkey verify route (`/api/auth/mobile/passkey/verify`) already verified in-process and was unaffected.

### DIFF
--- a/web/src/app/api/passkey/authenticate/verify-authentication/route.ts
+++ b/web/src/app/api/passkey/authenticate/verify-authentication/route.ts
@@ -3,181 +3,34 @@
  *
  * Verify the WebAuthn authentication response and return user info.
  * This is the second (final) step in the passkey authentication ceremony.
- * Called by the NextAuth passkey provider to verify the user.
+ *
+ * Note: the NextAuth passkey provider does NOT call this over HTTP — it calls
+ * `verifyPasskeyAuthentication()` directly in-process. This route remains for
+ * any direct/external callers and shares the same verification core.
  */
 
 import { NextRequest, NextResponse } from "next/server";
-import {
-  verifyAuthenticationResponse,
-  AuthenticationResponseJSON,
-  VerifiedAuthenticationResponse,
-  AuthenticatorTransport,
-} from "@simplewebauthn/server";
-import { prisma } from "@/lib/prisma";
-import {
-  verifyAndConsumeChallenge,
-  base64URLToBuffer,
-  bufferToBase64URL,
-} from "@/lib/webauthn-utils";
-import { rpID, expectedOrigin } from "@/lib/webauthn-config";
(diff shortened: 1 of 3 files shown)


malinmalliyawadu:
Addressed the test-coverage gap from the review: added `web/src/lib/passkey-auth.test.ts` with 9 unit tests covering the verification core (404 not-found, 400 invalid/expired challenge, 400 bad signature, 400 unverified, 403 cloned-credential counter, counter=0 authenticators, success path + counter update, and archived-user auto-reactivation) — `@simplewebauthn/server` is mocked as suggested. All 273 unit tests pass.

Deferred the optional suggestions (status-code constants, wrapping challenge-consume + counter-update in a transaction, defensive `passkey.user` null check) — the transaction change would require threading a tx context through `verifyAndConsumeChallenge` and the ordering is unchanged from the pre-refactor route, so out of scope for this hotfix.

PR #1006, opened 2026-06-18
feat(mobile-admin): restaurant filter + admin push for messages & approvals

malinmalliyawadu:
# Pull Request

## Description

Two related additions to the mobile **admin** section:

1. **Restaurant filter** on the admin screens (Shifts + Approvals), with a **shared** selection that applies across the section.
2. **Admin push notifications** for the two events that previously notified no one: new volunteer→team messages, and new pending shift signups (approval requests).

### Restaurant filter (Shifts + Approvals)
- Shared, in-memory selection (Zustand) applied across the admin hub, approvals, and tonight's shifts — pick a restaurant once and it sticks while moving between screens. `null` = all restaurants, resets on launch (safe default).
- Drives the **existing** server-side `?location=` params on the shifts/approvals endpoints; query keys are location-scoped.
- Reuses the volunteer shifts location-picker UI (pill + bottom sheet) for consistency.
- New mobile-authed `GET /api/mobile/admin/locations` supplies the picker list.

### Approval-request push — location-scoped, reuses the restaurant-manager page
- `notifyManagersOfPendingSignup` notifies the `RestaurantManager`s whose assigned `locations` include the shift's location **and** who have `receiveNotifications` enabled — the existing toggle on the restaurant-managers admin page (no new UI). Bell + push.
- Hooked into **both** the web (`/api/shifts/[id]/signup`) and mobile (`/api/mobile/shifts/[id]/signup`) signup routes, after auto-approval runs (only fires if the signup is still `PENDING`/`REGULAR_PENDING`).

### New-message push — opt-in, toggle on the screen
- Volunteer→team messages now also push to admins who opt in. **Push-only, no DB notification row**, preserving the existing "don't drown admins" design (admins still get the live web SSE event as before).
- Bell toggle in the mobile messages header, **off by default**, persisted to the account (`User.adminMessageNotifications`) via `GET/PUT /api/mobile/admin/messages/notify-preference`. Enabling it also requests OS permission + registers the push token so the toggle can't silently no-op.

### Plumbing
- New `SHIFT_SIGNUP_REQUEST` `NotificationType` value + `User.adminMessageNotifications` field — each with a migration (mirrors the existing enum-add migration pattern).
- Mobile push tap-routing: `/admin/shifts/:id` → approvals queue, `/admin/messages` → inbox.

## Type of Change

## Version Bump Required
`version:minor`

## Testing
- [x] Typechecked both `web/` and `mobile/` against a freshly regenerated Prisma client — **zero** errors from these changes (only two pre-existing, unrelated errors on `main`).
- [ ] Unit tests — none added yet (see notes).
- [ ] Manual / device testing — a real push round-trip needs a physical device + Expo and wasn't possible in this environment.

## Reviewer notes
- **Sanity check (confirmed):** the admin tab is gated for non-admins both client-side (`isAdmin` tab render + route guard) and **server-side** (every `/api/mobile/admin/*` route, including the new ones, uses `requireMobileAdmin` → 403).
- **Why the manager page, not the in-app filter:** the restaurant filter is per-device/ephemeral and can't drive server-side push. Approval pushes use the persisted `RestaurantManager` → location mapping instead, which already has a `receiveNotifications` toggle.
- **Recipients differ by design:** approvals → location-matched managers; messages → all opted-in admins (messages aren't tied to a restaurant).
- **Scope:** only the two interactive signup routes (volunteer self-signup) trigger approval pushes. Regular-volunteer auto-generated `REGULAR_PENDING` signups created elsewhere are not covered yet.
- **Migrations** couldn't be applied locally (no DB in the worktree); CI's Postgres will run them. The enum-add migration mirrors `20260421000001_add_shift_shortage_notification_type`.

### DIFF
--- a/mobile/app/(tabs)/admin.tsx
+++ b/mobile/app/(tabs)/admin.tsx
@@ -5,6 +5,7 @@ import React, { useCallback } from "react";
 import { Pressable, ScrollView, StyleSheet, Text, View } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 
+import { AdminLocationFilter } from "@/components/admin/location-filter";
 import { Eyebrow } from "@/components/ui/eyebrow";
 import { Brand, Colors, FontFamily, Palette } from "@/constants/theme";
 import { useColorScheme } from "@/hooks/use-color-scheme";
@@ -13,6 +14,7 @@ import {
   useAdminToday,
   useAdminUnreadCount,
 } from "@/hooks/use-admin";
+import { useAdminLocationFilter } from "@/lib/admin-location-filter";
 import { useAuth } from "@/lib/auth";
 
 /* ─── Editorial palette (mirrors the Help landing) ──────────────── */
@@ -52,9 +54,10 @@ export default function AdminScreen() {
   const router = useRouter();
   const { user } = useAuth();
 
+  const selectedLocation = useAdminLocationFilter((s) => s.selected);
   const unread = useAdminUnreadCount(true);
-  const pending = useAdminPending();
-  const today = useAdminToday("");
+  const pending = useAdminPending(selectedLocation);
+  const today = useAdminToday("", selectedLocation);
 
   const refetchAll = useCallback(() => {
     void pending.refetch();
@@ -118,6 +121,7 @@ export default function AdminScreen() {
 
         <View style={styles.channelEyebrow}>
           <Eyebrow color={paperTint.eyebrow}>What needs you</Eyebrow>
+          <AdminLocationFilter />
         </View>
 
         <View style={styles.optionList}>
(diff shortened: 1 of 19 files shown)


malinmalliyawadu:
Thanks for the review! Addressed the cheap "before merge" polish in 825e113a:

- 🔧 **Toggle save errors** now surface an `Alert` (was a silent optimistic rollback).
- 🔧 **Standardized fire-and-forget** — dropped the mixed `void` on the pending-signup notify call to match the existing `.catch()` pattern in the same signup routes.
- 💎 **Documented the no-badge decision** for the admin message push (admins track message unread via the thread's `teamLastReadAt`/SSE, not Notification rows, so there's no per-user count to surface — and we don't want to clobber their real notification badge).

Deferred (non-blocking, as you noted):
- 🧪 Unit tests for `notifyManagersOfPendingSignup` / `broadcastNewMessageToAdmins` — happy to add in a follow-up.
- Migration dates are final (real `20260618…` timestamps, not placeholders).

Typechecked both `web/` and `mobile/` after the changes — green.

PR #1007, opened 2026-06-18
test(notifications): admin signup-request + message push helpers (+ review polish)

malinmalliyawadu:
# Pull Request

## Description

Follow-up to #1006. Two parts:

1. **Unit tests** for the two notification helpers introduced in #1006 (the deferred item from that review).
2. **Re-applies the review polish** that didn't make it into #1006 — it was merged before that commit landed, so `main` is currently missing it.

### Tests (`web/`, Vitest)
- **`notifyManagersOfPendingSignup`** — notifies location-matched, opted-in managers with a bell notification + push for `PENDING`/`REGULAR_PENDING` signups; no-ops for auto-approved signups, a shift with no location, no assigned managers, and a deleted signup.
- **`broadcastNewMessageToAdmins`** — always broadcasts over SSE; pushes **only** to admins who opted in (`adminMessageNotifications`), **excluding the sender**; plus email-name fallback and long-body preview truncation.
- 11 tests, all passing locally.

### Re-applied polish (was lost from #1006)
- Surface an `Alert` if the message-notification toggle fails to save (was a silent optimistic rollback).
- Drop the mixed `void` on the fire-and-forget pending-signup notify call, matching the existing `.catch()` pattern in the same signup routes.
- Document why the admin new-message push intentionally carries no badge.

## Type of Change
- [x] 🧪 Tests (adding missing tests)
- [x] 🧹 Code refactoring (the re-applied polish)

## Version Bump Required
`version:patch`

## Testing
- [x] Unit tests pass (`vitest run` — 11/11 in the two new suites)
- [x] Typecheck clean for the new files

## Reviewer notes
- The polish re-application is a clean cherry-pick of the commit that missed the #1006 merge — verified `main` does not already contain it.
- Tests mock `@/lib/prisma` and the push/SSE dependencies per the repo's existing test conventions (e.g. `friend-request-service.test.ts`).

### DIFF
--- a/mobile/hooks/use-admin.ts
+++ b/mobile/hooks/use-admin.ts
@@ -1,3 +1,4 @@
+import { Alert } from "react-native";
 import {
   useMutation,
   useQuery,
@@ -119,9 +120,14 @@ export function useAdminMessageNotifyPref() {
       return { previous };
     },
     onError: (_err, _enabled, context) => {
+      // Roll the optimistic toggle back and let the admin know it didn't stick.
       queryClient.setQueryData(
         queryKeys.admin.messageNotifyPref(),
         context?.previous ?? false
+      );
+      Alert.alert(
+        "Couldn't update notifications",
+        "Please check your connection and try again."
       );
     },
     onSuccess: (enabled) => {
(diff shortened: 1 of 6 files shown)


malinmalliyawadu:
Thanks for the LGTM! Added the two optional coverage suggestions in 5d050c0f:

- ✅ **No-badge assertion** on the admin message push (`expect(payload.badge).toBeUndefined()`), pinning the documented intent that admins track message unread via `teamLastReadAt`, not Notification rows.
- ✅ **NZ date-formatting test** for `notifyManagersOfPendingSignup` — uses a `13:00 UTC` start that crosses midnight into the next NZ day, so a UTC-vs-NZ regression would surface as "30 Jun" instead of "1 Jul".

Both suites green: 12/12. `version:patch` label is on.

PR #1011, opened 2026-06-18
fix(push): surface silent Expo push failures for admin message notifications

malinmalliyawadu:
## Description

Admins who had the opt-in enabled (`adminMessageNotifications`) and a registered device still weren't receiving mobile push notifications when volunteers messaged the team — and the send path made the failure **invisible**, reporting success even when nothing was delivered.

This PR fixes three silent-failure paths in the Expo push service and adds the observability needed to pin down the root cause from production logs.

### What changed (`web/src/lib/services/expo-push.ts`)

1. **Top-level Expo errors were dropped.** The send response was read as `json.data ?? []`, so when Expo rejected the *whole request* (enhanced-security access token required, malformed payload), `data` was `undefined`, the loop did nothing, and the code reported success. Now `json.errors` is logged and counted.
2. **`InvalidCredentials` deleted valid tokens.** That code (and `MismatchSenderId`) means the project's **APNs/FCM push credentials** are misconfigured — not that the device is dead. The old code lumped it in with `DeviceNotRegistered` and deleted the token, permanently wiping a valid registration so every later send became a no-op. Now only `DeviceNotRegistered` deletes a token; credential errors are logged loudly and the token is kept.
3. **Delivery failures were never checked.** An accepted (`ok`) ticket only means Expo *queued* the push; APNs/FCM delivery errors surface in the **receipt**, which was never fetched. Added a fire-and-forget receipt check (15s later) that logs exactly why a push didn't land and cleans up genuinely dead tokens.

Also added:
- Optional **`EXPO_ACCESS_TOKEN`** support (Authorization header) for projects with *Enhanced Security for Push Notifications*.
- **Funnel logging** in `web/src/lib/messaging-notify.ts`: `volunteer X messaged team; N opted-in admin(s) to push`, plus a dispatch summary `N accepted, M error(s)`.

## Type of Change

## Version Bump Required
`version:patch`

## Testing
- [x] Typecheck passes for the changed files
- [ ] Manual testing: deploy, have a volunteer send a message, read Coolify logs for `[messaging-notify]` / `[EXPO_PUSH]` to confirm the delivery outcome

## Additional Notes

These changes don't *guess* the root cause — they make it visible in server logs. After deploy, a volunteer test message will log one of:
- `0 opted-in admin(s)` → recipient/opt-in side
- `Request rejected … errors` → enhanced security / needs `EXPO_ACCESS_TOKEN`
- `Delivery failed (InvalidCredentials/MismatchSenderId)` → **APNs/FCM credentials misconfigured** on the EAS project (leading suspect)
- `Delivery failed (DeviceNotRegistered)` → stale token

### DIFF
--- a/web/src/lib/messaging-notify.ts
+++ b/web/src/lib/messaging-notify.ts
@@ -112,6 +112,9 @@ export async function broadcastNewMessageToAdmins(
       },
       select: { id: true },
     });
+    console.log(
+      `[messaging-notify] volunteer ${args.volunteer.id} messaged team; ${optedInAdmins.length} opted-in admin(s) to push`
+    );
     if (optedInAdmins.length > 0) {
       await sendPushToUsers(
         optedInAdmins.map((a) => a.id),
--- a/web/src/lib/services/expo-push.ts
+++ b/web/src/lib/services/expo-push.ts
@@ -1,7 +1,28 @@
 import { prisma } from "@/lib/prisma";
 
 const EXPO_PUSH_URL = "https://exp.host/--/api/v2/push/send";
+const EXPO_RECEIPT_URL = "https://exp.host/--/api/v2/push/getReceipts";
 const MAX_CHUNK = 100; // Expo's per-request message limit
+
+/**
+ * Required when the Expo project has "Enhanced Security for Push Notifications"
+ * enabled — without it Expo rejects every send with a top-level error. Optional
+ * otherwise. Set EXPO_ACCESS_TOKEN in the server environment if pushes silently
+ * stop working after enabling enhanced security.
+ */
+const EXPO_ACCESS_TOKEN = process.env.EXPO_ACCESS_TOKEN;
+
+function expoHeaders(): Record<string, string> {
+  const headers: Record<string, string> = {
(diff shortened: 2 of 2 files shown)


malinmalliyawadu:
Thanks for the thorough review! Addressed the two actionable items in `2cae206c`:

- **Test coverage** — added `web/src/lib/services/expo-push.test.ts` covering the three regression-critical branches you flagged: `InvalidCredentials` keeps the token, `DeviceNotRegistered` deletes it, and top-level Expo errors are logged/counted instead of silently succeeding (plus invalid-token short-circuit and `isExpoPushToken` format checks). 5 tests, all passing.
- **Error-code constants** — extracted `EXPO_ERROR` to replace the repeated magic strings across the send and receipt paths.

Left as-is (with rationale):
- **15s receipt delay** — kept intentionally close to the triggering event for diagnostics; easy to bump if prod logs show missing receipts, as you noted.
- **Receipt-fetch chunking / double-delete race** — both are non-issues at the current admin-notification scale (small opted-in set), and `deleteMany` is idempotent. Worth revisiting only if this path ever fans out to 1000+ recipients.

PR #1019, opened 2026-06-18
feat(admin): sync service-night Stripe koha from the Stripe API

malinmalliyawadu:
## Description

Makes the Service Night Report's **Sync** button real. It pulls the night's Stripe koha for the selected location from the Stripe API and fills the **Stripe** field — non-destructive, so the admin reviews it and saves with the rest of the report. Previously the button was a "Soon" placeholder and the figure was typed in by hand.

### How location is attributed (the tricky bit)

EE runs a **single Stripe account**, so location can't be filtered server-side. It lives as a *string* in two flows, both handled by one normalised matcher (dash/case/space-insensitive):

1. **Donations** — a product per restaurant (`"You're supporting Everybody Eats – Wellington"`), reachable only via Checkout Session → line item → price → product (charges/PaymentIntents carry no product info, and `line_items` isn't expandable on the sessions *list* — fetched per session).
2. **Pay-at-table** (being built) — PaymentIntents with a `description` like `"Everybody Eats — Wellington"`.

A donation's Checkout Session and its underlying PaymentIntent are **deduped** so nothing is double-counted; non-NZD payments are ignored.

### Special-event / pop-up venues

Venues with `Location.isPopup = true` are **koha-manual-only**: the Sync button is hidden for them, and the sync route refuses them with a `422` backstop. Only the three real restaurants (Glen Innes / Onehunga / Wellington) sync.

### Graceful degradation

No real `STRIPE_SECRET_KEY` is wired yet. Until one is set, the button shows a friendly **"Stripe isn't connected yet"** message (the endpoint returns `503`). No DB schema change — attribution is by string match, nothing new to persist.

**Follow-up for the maintainer:** add a restricted, read-only `STRIPE_SECRET_KEY` (Checkout Sessions / PaymentIntents / Products read). Use an `sk_test_…` key first to validate against the test dashboard, and confirm the pay-at-table flow's `description` format matches `"Everybody Eats — <Location>"`.

## Type of Change

## Version Bump Required
`version:minor`

## Testing
- [x] Unit tests pass (350/350; 13 new in `stripe-koha.test.ts` covering matching, Special-events exclusion, session/PI dedupe, location split, NZD guard)
- [ ] E2E tests pass

Manual verification (preview, logged in as admin):
- No-key path: clicking **Sync** → `503` → "Stripe isn't connected yet" toast; field left untouched.
- Pop-up venue: temporarily set Wellington `isPopup=true` → Sync button hidden, Stripe field still editable; reverted → button returns.
- Auth guard: unauthenticated request → `403`.
- Typecheck + lint clean.

## Checklist
- [x] I have added tests that prove the feature works

## Additional Notes

New files: `web/src/lib/stripe.ts` (lazy client + `isStripeConfigured()`), `web/src/lib/stripe-koha.ts` (+ tests, pure helpers), `web/src/lib/services/stripe-koha-sync.ts` (Stripe fetch), `web/src/app/api/admin/meals-served/stripe-sync/route.ts` (ADMIN GET). `STRIPE_SECRET_KEY` added to `.env.example`. Sums gross of refunds for v1.

### DIFF
--- a/web/.env.example
+++ b/web/.env.example
@@ -55,6 +55,11 @@ CAMPAIGN_MONITOR_ANNOUNCEMENT_EMAIL_ID=""
 OPENROUTER_API_KEY=""
 OPENROUTER_MODEL="anthropic/claude-sonnet-4" # optional, this is the default
 
+# Stripe (service-night koha sync — optional; Sync button is disabled when unset)
+# Use a RESTRICTED, READ-ONLY key with read access to: Checkout Sessions,
+# PaymentIntents, and Products/Prices. Use a test key (sk_test_…) for local dev.
+STRIPE_SECRET_KEY=""
+
 # Supabase (for file storage)
 NEXT_PUBLIC_SUPABASE_URL=""
 NEXT_PUBLIC_SUPABASE_ANON_KEY=""
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -73,6 +73,7 @@
         "sanitize-html": "^2.17.5",
         "sharp": "^0.34.5",
         "sonner": "^2.0.7",
+        "stripe": "^22.2.2",
         "tailwind-merge": "^3.5.0",
         "tw-animate-css": "^1.4.0",
         "unpdf": "^1.6.2",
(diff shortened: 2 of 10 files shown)


malinmalliyawadu:
Thanks for the thorough review! Pushed `0de7254` addressing the highest-value item, with reasoning on the rest:

**Done**
- **Location existence (High):** unknown `location` now returns `404` before any Stripe call, instead of falling through to a `$0` sync.

**Intentionally not changed**
- **Currency null-safety:** `s.currency && s.currency.toLowerCase()` doesn't throw on `null` — the `&&` short-circuits. Switching to `?.` would also change semantics (a null-currency payment would flip from counted → skipped), so I kept the explicit guard.
- **Pagination safety cap:** for a koha *total*, a silent cap would make the figure wrong rather than just slow. A single restaurant-night won't approach hundreds of donations, so I'd rather be complete than bounded. Happy to add a logged hard-limit if volumes ever surprise us.
- **`maxDuration` / concurrency `p-limit` / route rate-limiting:** deferred — per-night session counts are small (tens), and `maxDuration` is Vercel-specific while this app is migrating off Vercel.
- **E2E test for the admin flow:** good follow-up; left for a future PR since it needs a mocked Stripe client.

Refund handling (gross for v1) and the pay-at-table `description` format remain noted in the PR description as follow-ups when a real key is wired.

PR #1022, opened 2026-06-20
fix(notifications): deep-link shift shortage notifications to the day

malinmalliyawadu:
## What changed

Shift shortage notifications now point volunteers to the **specific day** the shortage is for, instead of dumping them on the generic shifts page with a vague message.

A shortage send can cover multiple shifts, but it's **always for a single day**. Previously, multi-shift notifications:
- linked to `/shifts` (the location picker — no date context)
- said only *"Tap to see N shifts across our locations that still need cover."*

Now multi-shift notifications:
- **link to that date's shift list** — `/shifts/details?date=2026-02-08&location=Auckland` (the same deep-link pattern the shortage email already uses). Location is appended only when every shift shares one; otherwise just the date.
- **include the date** in the title and message, e.g.
  - Title: `3 shifts need volunteers on Saturday, February 8, 2026`
  - Message: `Saturday, February 8, 2026: 3 shifts at Auckland still need cover. Tap to help out.` (falls back to *"across our locations"* when shifts span multiple sites)

Single-shift notifications are **unchanged** — they already linked to `/shifts/{id}` and named the date.

## Why

Clicking a shortage notification took volunteers to the generic shifts page rather than the relevant date, and the message gave no indication of which day needed cover.

## Reviewer notes

- The multi-shift link drops the location query param when shifts span multiple locations, since `/shifts/details` filters to one location — the date alone still scopes it correctly.
- The deep-link target matches the `/shifts/details?date=...` link already used in the shortage **email**, so in-app/push and email now agree.

### DIFF
--- a/mobile/lib/notification-routing.ts
+++ b/mobile/lib/notification-routing.ts
@@ -77,6 +77,17 @@ export function navigateToNotificationTarget(actionUrl: unknown) {
     return;
   }
 
+  // /shifts/details?date=YYYY-MM-DD -> shifts tab focused on that day.
+  // Shortage notifications deep-link here; the shifts tab honours ?date=.
+  // (location isn't forwarded — the tab applies the user's own filter.)
+  if (pathname === "/shifts/details") {
+    const date = query.get("date");
+    router.push(
+      date ? { pathname: "/(tabs)/shifts", params: { date } } : "/(tabs)/shifts"
+    );
+    return;
+  }
+
   // /shifts or /shifts/mine -> shifts tab
   if (pathname === "/shifts" || pathname.startsWith("/shifts/")) {
     router.push("/(tabs)/shifts");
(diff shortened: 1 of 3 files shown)


malinmalliyawadu:
Thanks for the review. Addressed the Must Fix:

- **Email/push link inconsistency (🔴 Must Fix):** Fixed in `email-service.ts`. The aggregate `shiftsPageLink` now uses the same `allSameLocation` conditional as the push notification — it only appends `&location=` when every shift shares one location, otherwise links to the date only. Multi-location sends no longer wrongly filter the page to the first shift's location. (408b93e1)

Deliberately not changing:

- **Empty `shiftsForEmail[0]` guard (🟡):** In `route.ts` the `firstShift` access is already unreachable when empty — the handler returns 404 when `shifts.length === 0`, and the push block is gated behind `notifiedUserIds.length > 0` (which implies successful sends, which imply shifts). In `email-service.ts`, the route only ever calls it with a non-empty `shifts` array. No real null path.
- **Test coverage (🟡) / helper extraction (🟢):** Reasonable follow-ups, but out of scope for this focused fix; happy to do them separately if you'd like.

PR #1034, opened 2026-06-21
fix(locations): cascade location rename to free-text references

malinmalliyawadu:
## Description

Renaming a restaurant `Location` previously updated only the `Location` row. Because the location name is stored as a **free-text string (no foreign key)** across many tables, existing references kept the old name.

This surfaced as a **duplicate location in the mobile app** — e.g. both "Hopper" and "Hopper Cafe" appearing in the shift location filter. The mobile app derives its location list from distinct `Shift.location` strings ([`mobile/app/(tabs)/shifts.tsx`](https://github.com/everybody-eats-nz/volunteer-portal/blob/claude/focused-cohen-25909b/mobile/app/(tabs)/shifts.tsx)), while the web app reads the authoritative `Location` table ([`web/src/lib/locations.ts`](https://github.com/everybody-eats-nz/volunteer-portal/blob/claude/focused-cohen-25909b/web/src/lib/locations.ts)) — so the web app showed only the new name and mobile showed both.

## What changed

The admin `PATCH /api/admin/locations` handler now, **when the name actually changes**, runs a single transaction that:
1. Updates the `Location` row, and
2. Cascades the new name to every table storing the location as free text:
   - `Shift`, `ShiftTemplate`, `RegularVolunteer`, `AutoAcceptRule`, `MealsServed`, `DailyMenu`, `MessagingHours`
   - `User.defaultLocation`
   - `RestaurantManager.locations[]` (array element replaced in place via `array_replace`)

The transaction makes the rename atomic — if any update hits a unique constraint, the whole rename rolls back rather than leaving a half-applied state.

## Reviewer notes

- This prevents **future** renames from orphaning references. **Existing** stale rows (the current "Hopper" data) still need a one-off production cleanup — a corresponding SQL script renames `'Hopper'` → `'Hopper Cafe'` across the same tables and is being run separately against prod.
- `MealsServed`, `DailyMenu`, `ShiftTemplate`, and `MessagingHours` have unique constraints involving `location`; a real-world conflict during rename is practically impossible since the new name didn't exist as a `Location` before (which is uniquely constrained).

## Type of Change

## Testing
- [x] Typecheck passes for the edited file

## Additional Notes
Longer-term, normalizing `Shift.location` (and siblings) to a foreign key on `Location` would remove this class of bug entirely.

### DIFF
--- a/web/src/app/api/admin/locations/route.ts
+++ b/web/src/app/api/admin/locations/route.ts
@@ -2,6 +2,7 @@ import { NextRequest, NextResponse } from "next/server";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/lib/auth-options";
 import { prisma } from "@/lib/prisma";
+import { Prisma } from "@/generated/client";
 
 // Koha target per night — "" / null / undefined clear it; otherwise parse to 2dp.
 function parseTarget(value: unknown): number | null {
@@ -101,6 +102,16 @@ export async function PATCH(request: NextRequest) {
       );
     }
 
+    if (
+      name !== undefined &&
+      (typeof name !== "string" || name.trim() === "")
+    ) {
+      return NextResponse.json(
+        { error: "Location name must be a non-empty string" },
+        { status: 400 }
+      );
+    }
+
     const updateData: Record<string, string | number | boolean | null> = {};
     if (name !== undefined) updateData.name = name;
     if (address !== undefined) updateData.address = address;
@@ -136,6 +147,9 @@ export async function PATCH(request: NextRequest) {
       });
 
       if (isRename) {
+        // Only rows explicitly referencing the old name are touched; rows with
+        // a null location (templates/rules that apply to all locations) are
+        // intentionally left untouched.
         const where = { location: oldName };
         const data = { location: newName };
         await Promise.all([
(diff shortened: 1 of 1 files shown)


malinmalliyawadu:
Thanks for the thorough review. Pushed e68a2c0 addressing the actionable points:

- **Unique-constraint conflict (P2002)** — now caught and returned as a `409` with a clear message instead of a generic `500`. ✅
- **Input validation** — `PATCH` now rejects an empty/non-string `name` with a `400`. ✅
- **Null-location handling** — added a comment clarifying that rows with `location: null` (templates/rules that apply to all locations) are intentionally excluded from the cascade. ✅

Declined, with reasoning:

- **SQL injection in `$executeRaw` (flagged CRITICAL)** — this is a false positive. The code uses the **tagged-template** form (`` tx.$executeRaw`...` ``), which Prisma parameterizes — each `${oldName}`/`${newName}` becomes a bound placeholder (`$1`, `$2`, `$3`), not string interpolation. The injection-prone API is `$executeRawUnsafe`, which isn't used here. The suggested `Prisma.sql` rewrite is functionally identical, so no change needed.
- **`location` indexes** — renames are a rare admin action over volunteer-shift-sized data, so the added index write-cost isn't justified. Worth revisiting only if rename latency becomes a problem.
- **Dedicated rename e2e suite** — agreed it has value, but out of scope for this hotfix; noted as a follow-up alongside the longer-term FK normalization of `Shift.location`.

PR #1050, opened 2026-06-25
fix(analytics): use NZ calendar date for restaurant report bucketing & Service Nights table

malinmalliyawadu:
## Description
Fixes two related timezone bugs in the admin analytics restaurant reports, both rooted in the same storage detail: `MealsServed.date` is stored as **NZ-midnight expressed as a UTC instant** (e.g. a night on NZ calendar date **Jun 22** is stored as `2024-06-21T12:00:00Z`; **Jan 1** as `2023-12-31T11:00:00Z`). Reading the raw UTC components of that value lands a day (and sometimes a month/year) early.

### 1. Month/year bucketing (restaurant-reports.ts)
The report builder derived its month/year bucket keys from `rec.date.toISOString()`:

```ts
const iso = rec.date.toISOString();   // "2024-06-30T12:00:00.000Z"
const ym = iso.substring(0, 7);       // "2024-06"  ❌ should be 2024-07
const year = iso.substring(0, 4);     // Jan 1 → "2023" ❌ previous year
const weekday = toNZT(rec.date).getDay(); // NZT ✅ (already correct)
```

So **every night dated the 1st of a month was attributed to the previous month**, and **Jan 1 to the previous year**. Weekday bucketing already used NZT, so the handling was internally inconsistent. Fixed by deriving `year`, `ym`, and `weekday` all from `toNZT(rec.date)`.

### 2. Service Nights table date (service-nights/route.ts)
The per-night detail table emitted each row's date via `r.date.toISOString().substring(0, 10)`, so a night on **Jun 22** displayed as **Jun 21** (and likewise in the CSV export). Fixed by formatting the date in NZT (`formatInNZT(r.date, "yyyy-MM-dd")`).

## Type of Change

## Version Bump Required
`version:patch`

## Impact
Corrects in the analytics dashboard:
- Monthly & yearly trend series
- By-year and by-year-month tables
- New-volunteers-by-month series
- **Service Nights** detail table + CSV export date column

Not affected (and unchanged): the weekday `$/head` breakdown (already NZT) and the Summary by Location table (sums per location, no date attribution).

## Testing
- [x] Typecheck clean for both changed files (the only flagged lines are pre-existing worktree artifacts from the un-generated Prisma client typing `records` as `any`, unrelated to these changes)
- [ ] Manual browser verification not run — pure data-layer changes; this worktree has no DB/dev server. Behaviour is deterministic given the storage format documented above.

## Additional Notes
Separately worth hardening (not included here): the date-window bounds (`todayEnd`, custom `from`/`to`) in both the lib and this route are built with server-local `new Date(...)`, so they rely on the server process having `TZ=Pacific/Auckland`. On a non-NZ host this can drop boundary records. Happy to follow up if desired.

### DIFF
--- a/web/prisma/schema.prisma
+++ b/web/prisma/schema.prisma
@@ -771,7 +771,7 @@ model SiteSetting {
 // Meals served tracking (per day, per location)
 model MealsServed {
   id          String   @id @default(cuid())
-  date        DateTime // Date for this record (stored as start of day in UTC)
+  date        DateTime // Service night, stored as NZ-midnight expressed in UTC (e.g. NZ Jul 1 -> 2024-06-30T12:00Z). Read with toNZT()/formatInNZT() — a raw UTC slice lands a day early.
   location    String // Location name (Wellington, Glen Innes, Onehunga)
   mealsServed Int? // Customers / people served that day — headline count (nullable — admins may save notes without a count)
   notes       String? // Optional notes about the day
--- a/web/src/app/api/admin/analytics/service-nights/route.ts
+++ b/web/src/app/api/admin/analytics/service-nights/route.ts
@@ -2,7 +2,7 @@ import { NextRequest, NextResponse } from "next/server";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/lib/auth-options";
 import { prisma } from "@/lib/prisma";
-import { nowInNZT, toNZT } from "@/lib/timezone";
+import { formatInNZT, nowInNZT, toNZT } from "@/lib/timezone";
 import { parseDaysParam } from "@/lib/parse-days-param";
 
 const toNum = (v: unknown): number => {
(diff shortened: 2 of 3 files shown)


malinmalliyawadu:
Thanks for the thorough review! Pushed one follow-up in response (`44dcead2`): corrected the misleading `MealsServed.date` schema comment - it previously read "start of day in UTC" when the value is actually NZ-midnight expressed in UTC, which is exactly the non-obvious convention that led to these bugs. It now documents the format and points to `toNZT()`/`formatInNZT()`.

Deferring the other two recommendations as you suggested, since they're scoped as separate work:
- **Date-window boundary hardening** (`todayEnd`, custom `from`/`to` using server-local `new Date(...)`) - a behaviour change that deserves its own PR + verification, and only bites on a host without `TZ=Pacific/Auckland`.
- **Test coverage** (unit tests for bucketing, integration test for the service-nights date) - good follow-up; needs the DB/test harness this branch's worktree doesn't have.

PR #1052, opened 2026-06-27
fix(mobile): route multi-shift shortage notifications to shifts tab

malinmalliyawadu:
## Problem

Tapping a **shift-shortage notification** in the mobile app showed **"Shift not found"**.

Shortage notifications that cover **more than one shift on a day** deep-link to `/shifts/details?date=...` (set in `web/src/app/api/admin/notifications/send-shortage/route.ts`). In `mobile/lib/notification-routing.ts`, the generic `/shifts/:id` matcher (`^/shifts/([^/]+)$`) ran *before* the dedicated `/shifts/details` handler, so it captured `"details"` as a shift id and opened the shift-detail modal for a shift that doesn't exist → `mobile/app/shift/[id].tsx` renders "Shift not found". The real `/shifts/details` handler below was effectively dead code.

Single-shift shortage links (`/shifts/{realId}`) were unaffected, which is why this only bit multi-shift notifications.

## Fix

- Handle `/shifts/details` (with its `?date=` forwarding) **before** the generic `/shifts/:id` match.
- Exclude the non-id sub-route `mine` from the `:id` regex.

This mirrors the exact guard already present in the sibling `mobile/lib/deep-link-routing.ts`. The two mappers are meant to stay in sync (per its docstring) but had diverged - which is what caused this bug.

## Tests

- New `mobile/lib/notification-routing.test.ts` covering the multi-shift regression plus single-shift, `/shifts/mine`, dashboard, surveys, admin, and safety paths.
- `vitest run`: **28 passed** (12 new + 16 existing deep-link tests).
- Mobile `tsc --noEmit`: no type errors.

## Reviewer notes

- Mobile-only change; nothing observable in the web preview.
- Separate latent issue spotted while tracing (not in this PR): deleting a shift doesn't clean up its notifications, so a stale single-shift link can also 404. Flagged as a follow-up task.

### DIFF
--- a/mobile/lib/notification-routing.test.ts
+++ b/mobile/lib/notification-routing.test.ts
@@ -1,0 +1,105 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { navigateToNotificationTarget } from "@/lib/notification-routing";
+
+// `vi.mock` is hoisted above the import, so the module under test resolves
+// these stubs instead of the real expo-router / web-browser / store.
+const pushMock = vi.fn();
+const setPendingIdMock = vi.fn();
+const openBrowserMock = vi.fn();
+
+vi.mock("expo-router", () => ({ router: { push: (...args: unknown[]) => pushMock(...args) } }));
+vi.mock("expo-web-browser", () => ({
+  openBrowserAsync: (...args: unknown[]) => openBrowserMock(...args),
+  WebBrowserPresentationStyle: { AUTOMATIC: "AUTOMATIC" },
+}));
+vi.mock("@/hooks/use-pending-feed-item", () => ({
+  usePendingFeedItemStore: {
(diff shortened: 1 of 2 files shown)


malinmalliyawadu:
Thanks for the review! Addressed the two actionable suggestions in 37b5077f:

- **Array-based guard**: switched to `!["mine", "details"].includes(shiftDetail[1])` so the guard is now identical to `deep-link-routing.ts`, keeping the two mirrored mappers in sync.
- **Location-param test**: added a test documenting that `/shifts/details` forwards only `?date=` (the `&location=` the backend appends is intentionally dropped, since the mobile shifts tab applies the user's own location filter).

The orphaned-notification-on-shift-delete gap is already tracked as a separate follow-up. Tests: 10 passing in this suite; typecheck clean.

PR #1053, opened 2026-06-27
fix(shifts): delete dangling notifications when a shift is deleted

malinmalliyawadu:
## Description

Deleting a shift left behind notifications that deep-linked to it, so tapping them rendered **"Shift not found."** The `Notification` model has no FK/cascade to `Shift`, and neither admin DELETE handler cleaned up the notifications whose `actionUrl` pointed at the deleted shift.

This adds `deleteNotificationsForDeletedShifts()` and calls it inside the deletion transaction of both admin shift DELETE handlers (single shift and by-date).

### Why match on `actionUrl` rather than `relatedId`

Single-shift `SHIFT_SHORTAGE` notifications are created via `createNotificationsForUsers` **without** a `relatedId`, so `actionUrl` is the only reliable link back to the deleted shift. Matching on `actionUrl` also lets one helper cover every shift-linked notification type:

| actionUrl shape | Notification types |
|---|---|
| `/shifts/{id}` | `SHIFT_CONFIRMED`, `SHIFT_CANCELED`, single-shift `SHIFT_SHORTAGE` |
| `/admin/shifts/{id}` | `SHIFT_SIGNUP_REQUEST`, `SHIFT_CANCELLATION_MANAGER` |

Id-less links (`/shifts/mine`, `/shifts/details?date=...`) embed no shift id and are intentionally **not** matched.

## Changes

- **`src/lib/notifications.ts`** — new `deleteNotificationsForDeletedShifts(shiftIds, client?)`. Exact-matches the `actionUrl` deep links and `deleteMany`s them; accepts an optional transaction client; no-ops on an empty list.
- **`src/app/api/admin/shifts/[id]/route.ts`** — wrapped the previously-sequential signup + shift deletes in a `$transaction` and added the notification cleanup (signups → notifications → shift, atomic).
- **`src/app/api/admin/shifts/by-date/route.ts`** — added the cleanup inside the existing transaction.
- **`src/lib/delete-shift-notifications.test.ts`** — unit tests: both deep-link shapes, multi-shift, id-less links excluded, empty-list no-op, and that the passed transaction client is used.

## Type of Change

## Testing
- [x] Unit tests pass (`delete-shift-notifications.test.ts` 5/5, existing `notifications.test.ts` 7/7)
- [x] `tsc --noEmit` clean, ESLint clean on touched files
- [x] New tests added

This is a backend data-integrity fix; it's exercised by the unit tests rather than the browser preview.

## Additional Notes

Worth a follow-up: this is the classic missing-FK problem. A `relatedShiftId` column with `onDelete: SetNull` (or DB-level cleanup) would make this robust against future shift-linked notification types instead of relying on every deletion path calling the helper.


malinmalliyawadu:
Thanks for the review! The approval had only optional suggestions; I applied the two cheap, worthwhile ones in 466d68f4:

- **Exact-match documentation** — added a JSDoc note to `deleteNotificationsForDeletedShifts` clarifying it matches bare paths only, so a future `actionUrl` with a query string/fragment wouldn't be caught unless the helper is extended.
- **Cleanup logging** — both DELETE handlers now `console.info` the number of dangling notifications removed (`count > 0`) for production observability.

Left as-is (intentionally):
- **`@@index([actionUrl])`** — the review noted it's not necessary for the expected single-shift / by-date volumes; skipping to avoid an unused index.
- **`relatedShiftId` FK with `onDelete`** — the better long-term fix, already flagged as a follow-up in the PR description; out of scope for this bugfix.

Typecheck clean, ESLint clean, 12/12 unit tests pass.

PR #1065, opened 2026-06-28
test(e2e): fix flaky restaurant-manager Select interactions

malinmalliyawadu:
## What

Fixes the flaky `restaurant-manager-notifications.spec.ts` › **restaurant manager assignment workflow** e2e test (and hardens the rest of the spec).

## Why

The test failed once in CI then passed on rerun with no code change. Root cause is a classic Radix/shadcn `Select` timing race: after clicking the trigger the options render in a portal with an open animation, and the user list is fetched async. The test used a generic, un-scoped `page.locator('[role="option"]').first().click()`, which could resolve/click before the listbox was fully open and populated - timing out, and then failing the downstream form assertions. The same pattern was repeated throughout the file.

## Changes

- **New shared helper [`tests/e2e/helpers/select.ts`](web/tests/e2e/helpers/select.ts)** - there was no existing select helper (other specs inline the same fragile pattern). `selectOption`/`selectFirstOption`:
  - click the trigger, then wait for `getByRole("listbox")` to be **visible** (open animation done);
  - scope the option lookup to that open listbox so it can't match a stale/hidden portal;
  - wait for the option to be visible before clicking, then wait for the listbox to close.
- **Replaced every fragile `[role="option"]` pattern** in the spec with the helper.
- **Made the admin `describe` blocks idempotent** via `clearRestaurantManagerAssignments()` in `beforeEach`. While verifying with `--repeat-each` I found a second latent fragility: the tests pick "first admin / first location", so a pre-existing assignment flips the success toast to "Updated", flips the notifications checkbox to the stored value, and (with only ~3 selectable locations) eventually empties the dropdown. Clearing first makes every run a deterministic "create new" path - preserving the checkbox-default and "Successfully assigned" coverage rather than weakening it.

## Verification

Playwright, chromium:
- Target test `--repeat-each=5 --workers=1` on a fresh server: **5/5 pass**.
- Full file with the CI config (`--workers=2 --retries=2 --timeout=30000`): **all pass**. The Select interactions never failed once after the fix.
- `tsc --noEmit` and `eslint` clean on both files.

## Note for reviewers

This is unrelated to #1053 (notification-cleanup bugfix) - it is a pre-existing flaky test surfaced by that PR's CI run.

While running locally I observed occasional `page.goto: net::ERR_ABORTED`, traced to the app's SSE notification streams accumulating on a long-lived dev server. It is environmental (hits untouched tests too), not test logic, and CI's `retries: 2` + fresh-per-run server absorbs it.

### DIFF
--- a/web/tests/e2e/helpers/select.ts
+++ b/web/tests/e2e/helpers/select.ts
@@ -1,0 +1,60 @@
+import type { Locator } from "@playwright/test";
+import { expect } from "@playwright/test";
+
+/**
+ * Helpers for interacting with Radix/shadcn `<Select>` components in e2e tests.
+ *
+ * Radix renders the option list into a portal with an open animation, and the
+ * options themselves are frequently populated from an async fetch. A naive
+ * `trigger.click()` followed by `page.locator('[role="option"]').first().click()`
+ * is racy: the generic, un-scoped option locator can resolve against a portal
+ * that is still animating open (or a stale, hidden one), causing intermittent
+ * click timeouts. These helpers open the listbox, wait for it to be visible and
+ * populated, and scope the option lookup to that open listbox.
+ */
+
+interface SelectOptionTarget {
+  /** Pick the option at this index within the open listbox. Defaults to 0. */
+  index?: number;
+  /** Pick the first option whose text matches. Takes precedence over `index`. */
+  name?: string | RegExp;
+}
+
+/**
+ * Opens the Select identified by `trigger`, waits for the listbox portal to be
+ * open and populated, then clicks the requested option scoped to that listbox.
+ */
+export async function selectOption(
+  trigger: Locator,
+  target: SelectOptionTarget = {}
+): Promise<void> {
+  const page = trigger.page();
+
+  await trigger.click();
+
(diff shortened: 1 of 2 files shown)


malinmalliyawadu:
Thanks for the thorough review! Addressed the actionable nits in 31bb4816:

- **#4 `waitForTimeout(2000)` anti-pattern** — replaced both fixed delays (toggle/remove tests) with an assertion-based wait on the success toast, consistent with the workflow test.
- **#1 Empty-listbox edge case** — documented in `select.ts` JSDoc that the helper assumes a non-empty listbox and that callers who can hit an empty list should guard for it. (Didn't add a count assertion since `expect(options.first()).toBeVisible()` already serves as the populated-check, and `toHaveCount` takes an exact number rather than a min.)
- **#3 `!= null` nitpick** — applied; marginally more defensive against an explicit `null`.

**Deferred to follow-up** (out of scope here):

- **#2 Migrate other specs** — `achievements-leaderboard.spec.ts` (and a couple of others) still use the inline `[role="option"]` pattern. Left for a separate cleanup PR so this stays focused on the flaky test.

Re-verified: affected tests pass, `tsc`/`eslint` clean.

PR #1068, opened 2026-06-30
feat(admin): redesign volunteer availability card

malinmalliyawadu:
## Summary

Redesigns the **Availability & Preferences** card on the admin volunteer profile (`/admin/volunteers/[id]`) - it looked cluttered, and the "Available Days" badges appeared in an inconsistent order.

### Available Days
- New reusable `WeekAvailability` component renders a canonical **Monday-first week strip** (Mon → Sun), highlighting available days.
- **Fixes ordering structurally**: because it renders the full week and checks membership, the display is always Monday-first regardless of how the data was stored (lowercase `"monday"`, capitalised `"Monday"`, or out-of-order).
- Accessible: availability conveyed by **icon + colour** (not colour alone), with per-day `aria-label`s; plus an *"Available N days a week"* summary.
- Reused in both the regular-volunteer card and the main availability card for consistency.

### Available Locations
- Cleaner chips with a count summary (*"3 locations"*).
- The default location is highlighted inline with a ⭐ **Default** tag.
- Proper bordered empty state.

### Default Location
- Only shown as its own row when the default **isn't** already in the available list (avoids redundancy).
- Fixed the gap between the label and its value.

### How did they hear about us?
- Grouped into a tidy iconed row.

## Testing
- `npm run typecheck` and `eslint` pass on the changed files.
- Structure verified in-browser across all data scenarios (default-in-list, default-not-in-list edge case, single location, empty states); the Monday-first week strip verified styled in light and dark mode.

### DIFF
--- a/web/src/app/admin/volunteers/[id]/page.tsx
+++ b/web/src/app/admin/volunteers/[id]/page.tsx
@@ -32,6 +32,7 @@ import { cn } from "@/lib/utils";
 import { AdminPageWrapper } from "@/components/admin-page-wrapper";
 import { PageContainer } from "@/components/page-container";
 import { safeParseAvailability } from "@/lib/parse-availability";
+import { WeekAvailability } from "@/components/week-availability";
 import { VolunteerGradeToggle } from "@/components/volunteer-grade-toggle";
 import { VolunteerGradeBadge } from "@/components/volunteer-grade-badge";
 import { getDisplayGradeInfo } from "@/lib/volunteer-grades";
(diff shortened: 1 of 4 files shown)


malinmalliyawadu:
Thanks for the review! Addressed the two non-blocking suggestions in 18c250bc:

- **Unit tests**: extracted the Monday-first ordering/count logic into a pure `getWeekAvailability()` helper (`web/src/lib/week-availability.ts`) and added unit tests (`week-availability.test.ts`) covering ordering, counts, mixed case/whitespace, unknown values, and full/empty weeks. Extracting to a pure function keeps it testable under the project's `node`-environment vitest (no component-render infra needed). All 8 tests pass.
- **Null-safety**: the helper now guards against `null`/`undefined` input (`availableDays ?? []`).

PR #1069, opened 2026-07-01
feat(shifts): emoji + cover image support for Cafe Helpers shift type

malinmalliyawadu:
## What

Adds emoji and cover image support for a new **Cafe Helpers** shift type, on both web and mobile. Any shift type named `Cafe Helpers` (or whose name contains `cafe`/`café`/`barista`) now automatically resolves to a coffee theme - a ☕ emoji, a warm amber palette, and a cover image - through the existing name-keyed shift theme system.

No database seed is added: an admin creates the shift type through the existing flow and it picks up the theme by name.

## Why

We want a recognizable, welcoming identity for the new Cafe Helpers regular shift, consistent across the web portal and the mobile app.

## Changes

**Web**
- `shift-themes.ts`: added an optional, typed `coverImage` field to the theme, a `Cafe Helpers` exact entry (☕ + amber), and a `cafe`/`café`/`barista` keyword rule so location/variant names (e.g. "Cafe Helpers (Onehunga)") also resolve.
- Shift detail page: hero cover banner. Browse card: full-bleed cover strip. Both gated on `coverImage`, so every other shift type renders exactly as before.
- Cover asset: `web/public/shift-covers/cafe-helpers.webp` - the `wellington-dining-room` photo from the homepage hero carousel.

**Mobile**
- `mobile/lib/dummy-data.ts`: mirrors the ☕ emoji, amber colors, and the **same** carousel photo (converted to jpg) via `SHIFT_COVERS.cafeHelpers`, for both the exact and keyword theme entries - keeping web/mobile in sync per the existing cross-app parity requirement. Flows through the theme lookup the app already uses (hero image on shift detail, emoji + accent on cards).

**Tests**
- `shift-themes.test.ts`: covers the new exact/keyword resolution (including accented `café`) and cover-image presence/absence.

## Reviewer notes

- The cover only renders for shift types that define one, so this is purely additive - no visual change to existing types.
- The web card shows the cover **and** the small ☕ tile (the tile is kept for consistency with all other cards). Easy to drop the tile on cover cards if preferred.
- Web and mobile intentionally use the same photo in different formats (webp for web, jpg for mobile's asset convention).

## Verification

- Web: unit tests pass (40), `typecheck` clean, `lint` clean; cover render + crop confirmed in the hero and card.
- Mobile: unit tests pass (82); the changed file is `tsc`-clean.


malinmalliyawadu:
Thanks for the thorough review! 🙏 Addressed the actionable items in `0d54bb20`:

- **Keyword-priority test** (suggestion #2): added a test asserting `cafe` outranks `set up`/`service` in a combined name — `getShiftTheme("Cafe Set-up & Service")` → ☕. Locks in the rule order. (41 tests pass.)
- **webp/jpg divergence** (suggestion): documented the intentional split in the mobile `SHIFT_COVERS.cafeHelpers` comment (webp on web for size, jpg on mobile for the bundled-asset convention).
- **Bonus fix:** the mobile cover comment still said "Wellington dining room" — stale after the cover was swapped to a location-neutral photo. Corrected.

Intentionally skipped, with reasoning:
- **`width`/`height` on `` (CLS):** the covers already sit in space-reserving containers — `aspect-[3/1] sm:aspect-[4/1]` on the hero and a fixed `h-28` on the card — with the image `absolute/​h-full w-full object-cover`. The container reserves layout, so there's no shift to prevent; `width`/`height` on a fully-stretched image would be inert.
- **File-existence test:** the build already fails on a missing asset, and the path-string assertions cover the theme contract; a fs check would just couple the test to `/public` layout.
- **CLAUDE.md docs section:** happy to add one, but held off to avoid scope creep on this PR — can follow up if you'd like it.

PR #1070, opened 2026-07-01
feat(admin): shortage notifications analytics report with effectiveness drill-down

malinmalliyawadu:
# Pull Request

## Description
Adds a new admin analytics report at **`/admin/analytics/shortage-notifications`** that shows both the **volume** and the **effectiveness** of shift-shortage alerts, org-wide and per restaurant. It is built entirely on the existing `ShortageNotificationLog` history plus `Signup` records - no schema changes.

**What it answers**
- How many shortage notifications are we sending (send events and individual emails), across the org and at each site?
- How effective are they - i.e. how many volunteers actually signed up for a shift *after* being alerted about it?

**Key pieces**
- **Aggregation lib** (`src/lib/shortage-analytics.ts`): send events (batched by `sentAt`+`sentBy`) vs delivered emails, per-site breakdown (a multi-site send counts toward each site but once org-wide), monthly trend, and delivery success rate - all bucketed in NZ time.
- **Effectiveness / conversion**: joins each notified `(recipient, shiftId)` pair against signups and counts a conversion when the volunteer signed up **after** a delivered alert (`Signup.createdAt > sentAt`). Surfaced as a headline "Signups from Alerts" count and a per-site conversion rate. Denominator is delivered alerts; later cancellations still count (the nudge worked).
- **Drill-down modal**: lists the volunteers who signed up from an alert (avatar, the shift, timing e.g. "signed up 2 days later", link to the volunteer profile). Opens from the "Signups from Alerts" KPI card or any restaurant row, and is filterable by site.
- **Design**: follows the Restaurant Analytics dashboard - shared KPI cards with sparklines, StatChips, the `ChartCard` chassis, `InfoHint` popovers, and the shared chart theme.
- Wired into the admin nav under **Analytics**.

## Type of Change

## Version Bump Required
`version:minor` - new backward-compatible feature.

## Testing
- [x] Unit tests pass (17 tests in `shortage-analytics.test.ts` covering batch vs email counts, multi-site attribution, site filtering, delivery rate, NZ-timezone month bucketing, and conversion detection: before/after the alert, failed-email exclusion, denominator, per-site credit).
- [ ] E2E tests pass (not added in this PR)
- [x] Manual testing completed - verified the report, both metrics, the trend chart, the per-site table, and both drill-down modal entry points (org + per-site) in light and dark mode. Confirmed the filtered modal count matches the table (e.g. Wellington row = 7 signups -> modal shows 7).
- [x] New tests added

`npm run typecheck` and `npm run lint` are clean for the changed files.

## Checklist
- [x] I have added tests that prove the feature works

## Additional Notes
- No schema/migration changes - the report reads existing `ShortageNotificationLog` and `Signup` data.
- The conversion metric is a correlation proxy: it counts a signup that post-dates a delivered alert for the same shift. It excludes already-committed volunteers (their signup pre-dates the alert) but cannot prove strict causation.
- Headline "Signups from Alerts" counts delivered alerts that converted, while the modal lists distinct volunteer+shift signups; in normal use (one alert per person per shift) these are identical.

### DIFF
--- a/web/src/app/admin/analytics/shortage-notifications/page.tsx
+++ b/web/src/app/admin/analytics/shortage-notifications/page.tsx
@@ -1,0 +1,53 @@
+import { redirect } from "next/navigation";
+import { getServerSession } from "next-auth";
+import { authOptions } from "@/lib/auth-options";
+import { AdminPageWrapper } from "@/components/admin-page-wrapper";
+import { PageContainer } from "@/components/page-container";
+import { ShortageNotificationsAnalyticsClient } from "./shortage-notifications-analytics-client";
+import { LOCATIONS } from "@/lib/locations";
+import { getShortageNotificationAnalytics } from "@/lib/shortage-analytics";
+
+export default async function ShortageNotificationsAnalyticsPage({
+  searchParams,
+}: {
+  searchParams: Promise<{ [key: string]: string | string[] | undefined }>;
+}) {
+  const session = await getServerSession(authOptions);
+
+  if (!session?.user) {
+    redirect("/login?callbackUrl=/admin/analytics/shortage-notifications");
+  }
+
+  if (session.user.role !== "ADMIN") {
+    redirect("/dashboard");
+  }
+
+  const params = await searchParams;
+
+  const months = (params.months as string) || "12";
+  const location = (params.location as string) || "all";
+  const monthsNum = months === "all" ? 0 : parseInt(months, 10) || 12;
+
+  const data = await getShortageNotificationAnalytics(monthsNum, location);
+
+  const locationOptions = LOCATIONS.map((loc) => ({
+    value: loc,
+    label: loc,
+  }));
+
+  return (
+    <AdminPageWrapper
+      title="Shortage Notifications"
+      description="Shift-shortage alerts sent to volunteers and the signups they drove, org-wide and by restaurant"
+    >
+      <PageContainer testid="shortage-notifications-analytics-page">
+        <ShortageNotificationsAnalyticsClient
+          data={data}
+          months={months}
+          location={location}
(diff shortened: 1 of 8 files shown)


malinmalliyawadu:
Thanks for the thorough review! Pushed `428944cf` addressing the actionable low-priority items:

**Addressed**
- **Validate `months` param**: negative/invalid values now fall back to the 12-month default via a shared `parseMonthsParam` helper used by both API routes and the page (previously a `-3` slipped through to a future window). Added a unit test.
- **Correlation caveat in the UI**: the "Signups from Alerts" and "Conversion Rate" tooltips now state this is a correlation, not proof of cause.
- **Comment on fetching all signups**: documented why `getShortageConversions` intentionally includes every status (a later cancellation doesn't undo the nudge).
- **Year in modal date**: the conversions modal now shows the year when a shift isn't in the current year.

Test count is now 21 (added `parseMonthsParam` + the earlier per-location trend test); typecheck and lint are clean.

**Intentionally skipped (with reasoning)**
- **E2E test**: agreed it's valuable, but I'd rather add it as a focused follow-up than risk a flaky spec in this PR - noting it as a known gap.
- **Table-row `role="button"` semantics**: the current rows are keyboard-operable (tabIndex + Enter/Space) and you noted the approach works; reworking table semantics risks regressing that, so I left it.
- **Move `avgPerSend` server-side / reuse `getInitials`**: pure presentational derivations; keeping them client-side avoids churn without changing behavior.
- **Perf (pagination/indexing/caching)**: fair for very large datasets; the conversions list is already capped at 500 and volumes are admin-triggered (hundreds/month), so I'm leaving indexes/caching as a monitor-in-production follow-up rather than speculative optimization.

PR #1076, opened 2026-07-01
feat(admin): converters table + create notification groups from responders

malinmalliyawadu:
# Pull Request

## Description
Builds on the shortage-notifications report to add two connected things:

1. **A "Volunteers Who Convert" table** - one row per volunteer who signed up after a delivered alert (within the 3-day window), ranked by signups then conversion rate. Columns: alerts received, signups, personal conversion rate, last signup. Identifies your reliable responders.
2. **Create a notification group from them** - a checkbox column (select individual + select-all) and a "Create group" action that opens a name/description dialog and POSTs the selected volunteers to the **existing** `/api/admin/notification-groups`. The group is then immediately usable when sending shortage alerts (via the notifications page's "Load Group"), and stores the report filters as provenance.

No new group backend - it reuses the existing NotificationGroup system.

**Stacking:** this branch is stacked on `claude/shortage-3day-conversion-cap` (#1075) because the changes are intertwined in the same files. Base is set to that branch so this PR shows only the converters/groups diff; it will auto-retarget to `main` once #1075 merges.

## Type of Change

## Version Bump Required
`version:minor`

## Testing
- [x] Unit tests pass (26 total; +3 for the per-volunteer `aggregateConverters` rollup: stats/sort, window exclusion, per-site filtering)
- [x] Manual testing completed - verified live: the table lists the 33 converters (consistent with the report's "Signups from Alerts"), select-all + individual selection work, and creating a group produced an active NotificationGroup with 33 members that the send flow's members endpoint returns. Light + dark, no console errors.
- [x] New tests added

`typecheck` and `lint` clean.

## Additional Notes
- The converters rollup is fetched server-side in the report page (parallel with the existing analytics) and passed as a prop, so it re-computes on Apply Filters like the rest of the report.
- Group `filters` stores `{ source: "shortage-converters", months, location }` for provenance; the send page applies any recognised filter keys on load.

### DIFF
--- a/web/src/app/admin/analytics/shortage-notifications/create-notification-group-dialog.tsx
+++ b/web/src/app/admin/analytics/shortage-notifications/create-notification-group-dialog.tsx
@@ -1,0 +1,150 @@
+"use client";
+
+import { useState } from "react";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@/components/ui/dialog";
+import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
+import { Label } from "@/components/ui/label";
+import { Textarea } from "@/components/ui/textarea";
+import { Loader2, Users } from "lucide-react";
+import { toast } from "sonner";
+
+interface Props {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+  /** Volunteer IDs to add to the new group. */
+  memberIds: string[];
+  /** Stored on the group for provenance (e.g. report filters). */
+  filters?: Record<string, unknown>;
+  /** Prefilled group name. */
+  defaultName?: string;
+  /** Called after the group is created successfully. */
+  onCreated?: () => void;
+}
+
+export function CreateNotificationGroupDialog({
+  open,
+  onOpenChange,
+  memberIds,
+  filters,
+  defaultName = "",
+  onCreated,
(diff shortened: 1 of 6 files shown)


malinmalliyawadu:
Thanks for the review! Two things:

**Merge conflict resolved** — #1075 (the 3-day cap) squash-merged into `main`, so this branch still carried that commit as a separate SHA. Rebased onto `main` and dropped the now-redundant commit; the PR is back to a clean, mergeable state with only the converters/groups/home-site diff.

**Addressed**
- **Dialog reset** (`create-notification-group-dialog.tsx`): now clears the name field to `""` after a successful create instead of re-showing the just-used name (which would collide with the group just created). `29b4aca9`

**Intentionally deferred (with reasoning)**
- **API validation on `/api/admin/notification-groups`** (member IDs non-empty/unique, max name length): agreed, but that's the existing shared endpoint — out of scope here; worth a small follow-up PR so it applies to all callers (the existing "Manage Groups" flow too).
- **N+1 note on the 500-item `IN` query**: capped at 500 and admin-triggered volumes are small; Postgres handles a 500-id `IN` fine. Left a cap in place rather than adding pagination speculatively.
- **Zod validation of the `shifts` JSON**: the data is admin-generated and written by our own send path, so runtime schema validation would be defensive-only; noted as future robustness.
- **`useMemo` on `memberIds`**: harmless as the reviewer noted; left as-is.

Unit tests (26), typecheck, and lint are green on the rebased tree.

PR #1077, opened 2026-07-01
ci(test): don't let Playwright report publishing block the tests-passed gate

malinmalliyawadu:
## Description

The `tests-passed` required gate fails whenever the `merge-reports` job fails - but `merge-reports` only **publishes** the Playwright HTML report to Vercel and comments the link on the PR. That is informational, not a correctness signal.

On [#1053](https://github.com/everybody-eats-nz/volunteer-portal/pull/1053) all 20 Playwright shards passed, yet `merge-reports` failed because Vercel blocked the report-hosting team account:

```
Error: Your Team exceeded our fair use limits and has been blocked.
Learn More: http://vercel.link/fair-use
```

The empty deploy URL then made the PR-comment script exit 1, failing `merge-reports` and, through it, the required `tests-passed` check - blocking merges for a reason unrelated to test results. This affects every PR while Vercel is blocked.

## Change

Gate `tests-passed` on the actual `test` job only. `merge-reports` stays in `needs`, so the gate still waits for the report attempt and now echoes its status when it didn't succeed - but a failed/cancelled report no longer blocks the required check.

Real test failures still block merges (they surface via the `test` job, which is unchanged).

## Type of Change
- [x] 🔧 Build/CI changes

## Testing
- [x] YAML change is confined to the `run:` block of the `tests-passed` job; `test` job gating is unchanged
- [ ] Will confirm the gate goes green on this PR's own CI run

## Additional Notes

Root cause is a Vercel account/billing block on the `everybody-eats/volunteer-portal-playwright-reports` team (fair-use limit). This PR stops that from gating merges; separately, someone may want to unblock/upgrade that Vercel team so the HTML reports publish again.


malinmalliyawadu:
Thanks for the review - approving with optional suggestions only, no changes needed:

- **Notify the team when report publishing breaks** (Slack/Discord/email) - good idea, but it's a separate workflow rather than part of this gate change; leaving it out of scope here.
- **Resolve the Vercel fair-use block / alternative hosting** - not a code change; already called out in the PR description for whoever owns the Vercel account.
- **Drop `merge-reports` from `needs` entirely** - deliberately not doing this. As the review notes, keeping it in `needs` means the gate still waits for the publish attempt and surfaces its status, which is the intent.

Leaving the PR as-is.

Issue #10, opened 2025-08-13
Add e2e tests for Admin Dashboard page


malinmalliyawadu:
## Summary
Create comprehensive e2e tests for the admin dashboard page (`/admin`) to ensure admin users can access and navigate administrative functions.

## Test Coverage Needed
- [ ] Page loads for admin users only
- [ ] Admin authentication and role verification
- [ ] Dashboard statistics and metrics display
- [ ] Navigation to admin sub-sections
- [ ] Quick action buttons functionality
- [ ] Recent activity or summary information
- [ ] Admin role badge/indicator
- [ ] Unauthorized access prevention (non-admin users)
- [ ] Loading states for dashboard data
- [ ] Error handling for failed data loads
- [ ] Responsive design for admin interface

## Acceptance Criteria
- [ ] Only admin users can access the page
- [ ] Dashboard displays relevant admin information
- [ ] Navigation to admin sections works
- [ ] Proper role-based access control
- [ ] Tests use testid attributes
- [ ] Unauthorized access is properly blocked

## Priority
**Medium** - Admin functionality

## Implementation Notes
- Test with both admin and non-admin user sessions
- Verify role-based access control thoroughly
- Use testids for admin interface elements
- Test navigation to sub-admin pages

PR #1080, opened 2026-07-02
feat: quick location switcher with New badges, hide locations until shifts are published

malinmalliyawadu:
# Pull Request

## Description

Makes switching locations much easier on both the web and mobile apps, adds a subtle nudge when a new restaurant launches, and solves the "location created before its shifts exist" problem.

**The core concept:** a location doesn't exist for volunteers until it has upcoming shifts. New `Location.launchedAt` column (migration included, backfilled from `createdAt` for locations that already have shifts):

- A location created ahead of time stays **hidden** from all volunteer-facing pickers while it has no upcoming shifts. The admin locations page shows a "Hidden until shifts are published" badge so admins know why.
- When its first shifts are published, it appears automatically, `launchedAt` is stamped, and it's flagged **New** for 30 days.

**Web (`/shifts`)**
- The page title now doubles as the location switcher: the heading opens a dropdown of live locations, with an explicit "Change location" pill beside it (replaces the "← Choose Different Location" full-page round-trip).
- Menu shows a "New" badge next to recently launched locations, and the pill carries a green dot while any location is in its launch window.
- Logged-in users get a "Make X my default" action that saves to their profile.
- Fixes a latent bug: the page validated locations against a list loaded once at server start (top-level await in `lib/locations.ts`), so a newly created location wouldn't filter until a redeploy. The switcher list is now queried fresh per request via `getLiveLocations()` (React `cache()`-deduped).

**Mobile**
- The shifts-tab location filter is now persisted (Zustand + AsyncStorage, same pattern as the admin filter) so a switch survives app restarts instead of resetting to the profile default.
- `/api/mobile/shifts` now returns a server-driven `locations` list with `isNew` flags; the picker sheet shows NEW chips next to names and the location pill gets the green dot while a new location exists.

## Type of Change

## Version Bump Required
`version:minor` label added.

## Testing
- [x] Unit tests pass (423, incl. 5 new tests for `getLiveLocations`: hiding, launch-stamping, 30-day window, disabled locations, ad-hoc venues)
- [x] E2E tests updated (`shifts-browse.spec.ts` rewritten from the back-button flow to the switcher, incl. two new switching tests) — will run in CI
- [x] Manual testing completed (preview walkthrough: switching, badges, dot, chooser screen, mobile viewport; simulated a shift-less location staying hidden then launching with a New badge)

## Checklist

## Additional Notes
- Typecheck clean on web and mobile; lint has no findings in changed files.
- Seeds set `launchedAt` for the three established restaurants so fresh dev/CI databases don't badge them as New.
- The subtle-nudge approach is deliberate: no notifications. For a louder push admins can still use Announcements with location targeting.

### DIFF
--- a/mobile/app/(tabs)/shifts.tsx
+++ b/mobile/app/(tabs)/shifts.tsx
@@ -43,6 +43,7 @@ import {
   getShiftPeriodKey,
   getShiftPeriodLabel,
 } from "@/lib/shift-eligibility";
+import { useVolunteerLocationFilter } from "@/lib/volunteer-location-filter";
 import { getShiftThemeByName, type Shift } from "@/lib/dummy-data";
 
 /* ── Types ── */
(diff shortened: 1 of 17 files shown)


malinmalliyawadu:
Addressed the review in 02f78943:

**Fixed**
- **#1 Race condition in launch stamping**: `updateMany` now includes `launchedAt: null` in the WHERE, so concurrent requests can't re-stamp an already-launched location (unit test expectation updated).
- **#3 Whitespace normalization**: `Location` row names are now normalized the same way as shift venue strings when building the lookup map, so a name with stray spaces still matches (new unit test covers it).
- **#8 Error logging**: `setAsDefault` now logs the underlying error before showing the toast.
- **Coverage gap (mobile filter persistence)**: added `mobile/lib/volunteer-location-filter.test.ts` - 5 tests covering persistence (incl. explicit "All locations"), hydration, not clobbering a pre-hydration pick, and corrupted storage.

**Not changed, with reasons**
- **#2 Mobile API typing**: the client already types `locations?: BrowsableLocation[]` and defensive-defaults to `[]`; runtime Zod validation of API responses isn't a pattern anywhere in the mobile client, so I kept it consistent.
- **#5 Timezone**: `launchedAt > now - 30d` compares UTC instants of a rolling window; the result is identical regardless of display timezone, so `TZDate` wouldn't change behavior.
- **#7 Mobile hydration race**: `hydrate()` is invoked from a single effect, and a double-run (e.g. StrictMode) is idempotent - both reads produce the same `set`, and the `!state.hasChosen` guard protects user picks.
- **#9 e2e `waitForTimeout(500)`**: pre-existing code in a test this PR doesn't touch.
- **#10 aria-expanded**: Radix's `DropdownMenuTrigger` sets it automatically - verified in the browser (`aria-expanded` toggles on the trigger).
- **#4 configurable window / #6 switcher loading state**: agreed as possible follow-ups; leaving out of this PR.

All suites green: web 424 unit tests, mobile 87, typecheck clean on both.

PR #1082, opened 2026-07-05
feat: redesign friend profile pages with achievements on web and mobile

malinmalliyawadu:
## What changed

Complete visual overhaul of the friend profile pages on **web** and **mobile**, rebuilt around one shared information architecture that reads top to bottom as a story:

1. **Identity** — forest hero band, Fraunces display name, connected-since meta, relationship badges
2. **Together** — asymmetric bento with *shifts together* as the hero stat (dark forest panel, sun numeral), plus days connected, **mutual friends**, their hours and shifts
3. **Trophy shelf** *(new)* — the friend's achievements as round "plates": a featured latest-unlock card, a shelf of emoji plates with points, and an "X of Y unlocked · Z points" summary
4. **Their mahi** — rhythm strip (this month / avg / all-time) + favourite role
5. **Shared moments** — date-rail timeline with *Coming up* tags
6. **Join {name}** — their upcoming shifts; on web these now deep-link to `/shifts/details?date=…&location=…` instead of the generic shifts page

### Web
- `friends/[friendId]/friend-profile-content.tsx` fully rebuilt in the brand system (cream/forest/sun + Fraunces), replacing the old blue/purple/red gradient cards; matching Suspense skeleton in `page.tsx`
- Adds achievements and mutual-friends queries (mutual friends wasn't computed on web at all before)
- Hero sun disc softens to a glow below `sm` so it never sits behind the name

### Mobile
- `app/user/[id].tsx` friend view gains the Trophy shelf (featured unlock + horizontal plate shelf)
- Bento now shows the previously unused mutual-friends stat instead of duplicating total shifts (already shown as *All-time* in the rhythm strip)

### API
- `/api/mobile/users/[id]` friends-only payload now includes `achievements` (unlocked/total counts, points, 12 most recent items); types updated in `use-user-profile.ts`

## Why

The web page predated the brand refresh and was visually off-brand; achievements — a core part of the volunteer experience — were missing entirely from friend profiles on both platforms.

## Reviewer notes

- Achievements are only exposed in the accepted-friends payload/page, matching the existing visibility gating
- No e2e specs cover the friend profile page, so no test updates were needed; `friends.spec.ts` (list page) is untouched
- Verified in browser at desktop/375px/dark mode with seeded data; mobile API payload verified end-to-end with a real JWT; web lint + typecheck and mobile tsc clean

### DIFF
--- a/mobile/app/user/[id].tsx
+++ b/mobile/app/user/[id].tsx
@@ -777,8 +777,8 @@ function FullProfile({
         </View>
       </View>
 
-      {/* ── Trophy shelf ───────────────────────────────── */}
-      <SectionHeading title="Trophy shelf" colors={colors} />
+      {/* ── Achievements ───────────────────────────────── */}
+      <SectionHeading title="Achievements" colors={colors} />
       <View style={styles.section}>
         <TrophyShelf
           achievements={connection.achievements}
@@ -1279,9 +1279,11 @@ function TrophyShelf({
                   styles.plate,
                   {
                     backgroundColor: isDark
-                      ? "rgba(255,255,255,0.06)"
+                      ? "rgba(248,251,105,0.10)"
                       : Palette.cream100,
-                    borderColor: colors.border,
+                    borderColor: isDark
+                      ? "rgba(248,251,105,0.16)"
+                      : colors.border,
                   },
                 ]}
               >
--- a/mobile/app/user/[id].tsx
+++ b/mobile/app/user/[id].tsx
@@ -861,8 +861,8 @@ function FullProfile({
         </View>
       </View>
 
-      {/* ── Shared moments ─────────────────────────────── */}
-      <SectionHeading title="Shared moments" colors={colors} />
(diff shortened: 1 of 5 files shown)


malinmalliyawadu:
Addressed the review feedback in 6e487043:

- **P1 — mobile refresh error handling**: if the profile refresh fails right after accepting a friend request, the user now gets a "You're connected!" alert with a pull-to-refresh hint instead of a silent failure (the accept itself has already succeeded at that point).
- **P2 — `take: 12` on the achievements query**: intentionally *not* applied — `unlockedCount` and `totalPoints` are computed from the full result set, so capping the query would silently wrong those totals. Documented this in the route; the catalogue is <20 rows so the query stays cheap.
- **P2 — shared-shift matching**: documented that day/period/location matching is deliberately coarser than shift-ID matching so different roles worked during the same service still count as volunteering together.
- **P3 — e2e coverage**: deferred to a follow-up as suggested.

PR #1096, opened 2026-07-07
feat(chat-guides): replace volunteer grade with completed shift count

malinmalliyawadu:
## Summary

Refines the per-user context sent to the AI chat assistant to use real activity data instead of a grade label.

- **`preview/route.ts`** - removed the `Grade: … (GREEN/YELLOW/PINK)` line from the "About This Volunteer" context and replaced it with `Shifts completed: N`. The count uses the codebase's canonical definition (`CONFIRMED` signups on shifts that have already ended, matching `achievement-calculator.ts`). Dropped the now-unused `volunteerGrade` select.
- **`config/route.ts`** - removed the "What are the volunteer grades?" starter-question chip from the default suggested questions.

## Notes

- This touches the **admin chat preview** per-user context. The live mobile chat route sends no per-user data yet, so this does not change what volunteers currently receive - follow-up if we want mobile parity.
- The suggested-questions change only affects the **default**; if `CHAT_SUGGESTED_QUESTIONS` site setting is customized, that value still wins and should be edited in admin.

## Test plan

- [x] `tsc --noEmit` clean
- [ ] Admin chat preview shows "Shifts completed" and no grade
- [ ] Mobile chat suggested questions no longer include the grades chip

### DIFF
--- a/web/src/app/api/admin/chat-guides/preview/route.ts
+++ b/web/src/app/api/admin/chat-guides/preview/route.ts
@@ -54,7 +54,7 @@ export async function POST(request: Request) {
     const thirtyDaysAgo = new Date(Date.now() - 30 * 24 * 60 * 60 * 1000);
 
     // Fetch all context in parallel
-    const [resources, promptSetting, volunteer, upcomingShifts, locations, shiftTypes, achievements, totalVolunteers, recentMeals] =
+    const [resources, promptSetting, volunteer, upcomingShifts, locations, shiftTypes, achievements, totalVolunteers, recentMeals, completedShiftsCount] =
       await Promise.all([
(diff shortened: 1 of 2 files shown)


malinmalliyawadu:
Addressed the review's `completedShiftAdjustment` suggestion — it's a real correctness fix, not just consistency: the canonical total (`achievements.ts:343`, `auto-accept-rules.ts:69`) adds `completedShiftAdjustment` to the raw CONFIRMED-past count, so without it a volunteer with legacy adjustments would see a lower shift count in chat than on their profile. Now `Shifts completed` = raw count + adjustment. Pushed in b-fix commit. Typecheck clean.

PR #1097, opened 2026-07-07
feat(chat-guides): AI model config, preview override & content refine

malinmalliyawadu:
## Summary

Lets admins choose which OpenRouter model powers the chat assistant from the **Chat Guides** admin settings, instead of relying on the `OPENROUTER_MODEL` env var (which recently caused the assistant to run on `openai/gpt-5.4-nano` and fail to answer from the knowledge base).

- **New `CHAT_MODEL` site setting**, saved via the existing chat-guides settings PATCH endpoint.
- **Model resolution** in both routes: `CHAT_MODEL` setting → `OPENROUTER_MODEL` env → default `anthropic/claude-sonnet-4`. So the env var still works as a fallback, and nothing breaks if the setting is unset.
- **Admin UI**: an "AI Model" card with a free-text field (any OpenRouter slug) + recommended-model quick-pick chips, and a note steering away from underpowered nano/mini tiers.

## Files

- `api/mobile/chat/route.ts` - resolve + cache `modelId` in the shared static context.
- `api/admin/chat-guides/preview/route.ts` - resolve `modelId` from the setting.
- `api/admin/chat-guides/settings/route.ts` - accept + upsert `CHAT_MODEL`.
- `admin/chat-guides/page.tsx` - load the setting, pass to the client.
- `admin/chat-guides/chat-guides-content.tsx` - the "AI Model" card.

## Notes

- Chose a **free-text field + quick-pick chips** over a hardcoded dropdown deliberately: model slugs change over time and a fixed list would go stale. The chips give guardrails against the nano footgun without the maintenance burden.
- The mobile route caches the model in its 5-min static context, so a model change takes effect on the next cache refresh (or app restart) - same as the system prompt.
- Also cleaned up two pre-existing lint warnings in the touched file (unused `Check` import, unused catch binding).

## Test plan

- [x] `tsc --noEmit` clean
- [x] `eslint` clean on changed files
- [ ] Admin sets a model in Chat Guides settings → saved to `CHAT_MODEL`
- [ ] Chat assistant (preview + mobile) uses the configured model; falls back to default when blank

> Note: the admin Chat Guides page is auth + `CHAT_GUIDES` feature-flag gated, so I verified via typecheck/lint rather than driving the UI end-to-end in this environment.

### DIFF
--- a/web/src/app/admin/chat-guides/chat-guides-content.tsx
+++ b/web/src/app/admin/chat-guides/chat-guides-content.tsx
@@ -7,7 +7,6 @@ import {
   Trash2,
   Pencil,
   FileText,
-  Check,
   X,
   Info,
   Send,
@@ -93,6 +92,7 @@ interface ChatGuidesContentProps {
   estimatedTokens: number;
   initialSystemPrompt: string;
   initialSuggestedQuestions: string; // JSON string
+  initialModel: string;
 }
 
 const CATEGORY_LABELS: Record<string, string> = {
@@ -132,12 +132,25 @@ const DEFAULT_QUESTIONS: SuggestedQuestion[] = [
   { emoji: "📍", label: "Where are the kitchens?" },
 ];
 
+// Fallback model used when no CHAT_MODEL setting is saved (must match the API routes).
+const DEFAULT_MODEL = "anthropic/claude-sonnet-4";
+
+// Recommended models — capable enough to answer reliably from the knowledge base.
+// Avoid "nano"/"mini" tiers: they struggle to retrieve facts from long context.
+const RECOMMENDED_MODELS = [
+  "anthropic/claude-sonnet-4",
+  "anthropic/claude-3.7-sonnet",
+  "openai/gpt-5",
+  "google/gemini-2.5-flash",
+];
+
 export function ChatGuidesContent({
   initialChatResources,
   allResources,
   estimatedTokens: initialTokens,
   initialSystemPrompt,
   initialSuggestedQuestions,
+  initialModel,
 }: ChatGuidesContentProps) {
   const { toast } = useToast();
   const [chatResources, setChatResources] = useState(initialChatResources);
@@ -155,6 +168,7 @@ export function ChatGuidesContent({
       return DEFAULT_QUESTIONS;
     }
   });
+  const [model, setModel] = useState(initialModel);
   const [isSavingSettings, setIsSavingSettings] = useState(false);
 
   // Add resource dialog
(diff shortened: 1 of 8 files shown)


malinmalliyawadu:
Expanded this PR beyond the original model-setting with two related chat-guides authoring improvements:

**Per-preview model override** — the admin Chat Preview panel now has a Model field (preset dropdown via `datalist` + free text). It sends the model as a per-request override so admins can A/B a model without changing the saved `CHAT_MODEL` setting. Resolution: request override → `CHAT_MODEL` → `OPENROUTER_MODEL` env → default.

**"Refine with AI"** — new admin-only `POST /api/admin/chat-guides/refine` endpoint sends raw scraped/imported text to the configured model and returns cleaned-up markdown (strips nav/boilerplate/artifacts, preserves all facts). Buttons added to the edit, import-from-URL, and create-guide dialogs. Uses the same model resolution as the chat routes.

Typecheck + eslint clean. As before, the admin page is auth + `CHAT_GUIDES` flag gated, so verified via typecheck/lint rather than driving the UI end-to-end.

_🤖 via [Claude Code](https://claude.com/claude-code)_

malinmalliyawadu:
Resolved the merge conflict with `main` (#1096's shift-count change to the same preview route — combined both destructured vars) and addressed the review:

**Applied**
- **Model validation (#1)** + **trim (#2)** — settings PATCH now trims and validates the model id. Used a *permissive* check (`isValidChatModelId`) that allows tagged OpenRouter slugs like `anthropic/claude-3.5-sonnet:beta`. I deliberately did **not** use the suggested `^[a-z0-9-]+/[a-z0-9.-]+$` regex — it rejects the `:free`/`:beta`/`:nitro` tags OpenRouter actually uses, which would block valid models.
- **Tests + dedupe (#5)** — extracted the fallback chain into `lib/chat-model.ts` (`resolveChatModel`), now the single source of truth for the mobile, preview, and refine routes, with 7 unit tests covering resolution order, blank/whitespace handling, and validation.

**Skipped, with reasoning**
- **Health-check endpoint (#4)** — the Chat Preview panel (with the per-preview model picker) already lets admins live-test a model before relying on it, so a separate validation endpoint would be redundant.
- **Client-side real-time validation (#3)** — the PATCH now returns a clear error toast on save; a live validator on a trusted admin-only field isn't worth the complexity.
- **Migration docs (#6)** — the PR description already documents that `OPENROUTER_MODEL` keeps working as a fallback.

tsc + eslint clean; `vitest run src/lib/chat-model.test.ts` → 7 passed.

PR #1103, opened 2026-07-08
fix(sse): time out stalled SSE writes and decouple admin push from SSE broadcast

malinmalliyawadu:
## Root cause (confirmed in prod via #1102's tracing)

A volunteer test send at 06:54 logged:

```
[messaging] message … persisted …
[messaging] queueing admin fan-out …
[messaging-notify] SSE admin broadcast starting …
```

…and then nothing. `broadcastNewMessageToAdmins` awaits the SSE admin broadcast before the push fan-out, and `writer.write()` on a raw `TransformStream` to a client that vanished without a TCP reset never settles — it stalls forever on backpressure (the admin had 3 rapidly re-created SSE connections; at least one was a zombie). Everything behind that await — the opted-in-admin query, the funnel log, and the Expo push — was silently starved. This is why admins get no push when a volunteer messages the team, with zero error logs.

## Changes

- **`notification-sse-manager.ts`**: each SSE write races a 10s timeout. On expiry the writer is aborted and the connection dropped, so one dead client can no longer wedge the write loop (which also serves the bell badge and admin inbox live updates).
- **`messaging-notify.ts`**: the SSE broadcast and the admin push now run concurrently — the push path no longer waits on SSE at all.
- **`notification-sse-manager.test.ts`** (new): covers healthy delivery, timeout→abort→drop, healthy connection behind a stalled one, and admin broadcast completing despite a stalled connection.

## Testing

- `npm run typecheck` clean, eslint clean on touched files
- New SSE manager tests: 4/4 pass (fake timers)
- `messaging-notify` + `expo-push` suites: 11/11 pass

After deploy, a volunteer test send should produce the full 8-line chain from #1102, and a zombie connection now surfaces as `[SSE] Write to user … timed out after 10000ms … — dropping connection` instead of infinite silence.


malinmalliyawadu:
Addressed review feedback in 2a40185f: adopted suggestion 3 — a user's SSE connections are now written concurrently, so multiple zombie connections cost one 10s timeout window instead of one each (matches the observed 3-connection reconnect burst), with a test pinning that behaviour. The other suggestions (abort-error logging, extra edge-case test, log wording, env-configurable timeout) were left as-is per the review's own assessment that they're optional and current behaviour is correct.

PR #1104, opened 2026-07-08
feat(admin): bulk delete shifts by date range

malinmalliyawadu:
# Pull Request

## Description
Adds the requested ability to delete a whole date range of shifts at once, instead of deleting day by day (e.g. 4 months of wrongly-entered shifts).

- New `GET`/`DELETE /api/admin/shifts/by-date-range` route (admin-only): `GET` previews the blast radius (shift count, signed-up volunteer count, shift types) and `DELETE` removes signups, dangling notifications, and shifts in one transaction — mirroring the existing `by-date` route. Validates date format and order, and caps ranges at 366 days.
- New **Delete Shifts by Date Range** dialog with the same range calendar used by bulk shift creation, a live preview of what will be deleted, and warnings when volunteers are signed up or the range includes past days (attendance history). The confirm button stays disabled until the range contains shifts and shows the exact count ("Delete 30 Shifts").
- The toolbar's standalone **Delete All** button is replaced by a single **Delete** dropdown with "All shifts for this day" (disabled when the day has no shifts) and "Shifts in a date range...". Both dialogs now support controlled open state so the menu can drive them.

## Type of Change

## Version Bump Required
`version:minor`

## Testing
- [x] Unit tests pass
- [x] E2E tests pass

Extended `admin-shift-bulk-delete.spec.ts`: dialog flow across a range (create shifts on two days → pick range in the calendar → preview shows 2 shifts → delete → success banner), API validation (backwards range, >366 days, missing location), location scoping (other location's shifts survive), and volunteer 403 authz. The range tests use their own location + ~90-day-out window so they can't interfere with the single-day tests under parallel workers.

Manually verified in the browser: preview counts against real data (30 shifts / 48 volunteers across 8 days), full create→preview→delete→verify API cycle, disabled menu item on empty days, and dark-mode dialog styling.

## Additional Notes
Like the existing single-day delete, this does not email affected volunteers — the dialog warns the admin to notify them first. Happy to add cancellation notifications as a follow-up if wanted.


malinmalliyawadu:
Addressed the review in 40f9eb33:

- **#3 (shared statuses)**: Extracted `ACTIVE_SIGNUP_STATUSES` to `src/lib/signup-constants.ts`, used by both the `by-date` and `by-date-range` routes.
- **#5 (spinner)**: The delete-all dialog now uses `Loader2Icon` like the range dialog.
- **#6 (disabled item)**: The day-delete menu item shows a "(none)" hint when the day has no shifts.

Not taken, with reasons:

- **#1 (N+1)**: Prisma compiles nested relation `select`s into a join/batched query, not per-row queries — no N+1 here.
- **#2 (empty date bypass)**: `""` is falsy, so `!startDate` catches the empty string before the regex; the suggested bypass can't occur.
- **#4 (preview 404)**: The preview `GET` never returns 404 — it returns `{ shiftCount: 0 }` with 200, and the dialog renders the empty state. Only `DELETE` 404s, which via the UI can only happen if the shifts vanished between preview and confirm, where a visible error is the right feedback.
- **#7 (redirect params)**: The page only uses `date`/`location`; copying all current params would re-trigger stale banner params like `created=`. This matches the existing delete-all behaviour.
- **#8/#9 (test date math / year boundary)**: `setDate` is calendar-safe across leap years and month rollovers and matches the existing `getUniqueTestDate` helper; the range logic has no year-specific code paths that a Dec→Jan test would exercise.

PR #1105, opened 2026-07-08
feat(admin): email verification status and actions on volunteer profile

malinmalliyawadu:
# Pull Request

## Description
Volunteers with an unverified email are blocked from shift signup by the `emailVerified` gate, but admins had no UI to see or fix this - the only recourse was raw DB access.

This PR adds, on the admin volunteer profile page (`/admin/volunteers/[id]`):

- **Status badge**: "Email Verified" (emerald) or "Email Not Verified" (amber) alongside the existing status badges.
- **Email Verification section** in the Admin Actions card, shown only while the email is unverified, explaining that the volunteer cannot sign up for shifts and offering two actions:
  - **Resend verification email** - new admin-authenticated endpoint `POST /api/admin/users/[id]/email-verification`. Unlike the public `/api/auth/resend-verification`, it skips `checkForBot` bot protection and returns real errors instead of anti-enumeration responses.
  - **Mark as verified** - `PATCH` on the same route, behind a confirmation dialog with an optional reason. It sets `emailVerified`, clears any pending verification token, and **always creates an AdminNote** recording the override (plus the reason if given), so there is an audit trail.

## Type of Change

## Version Bump Required
`version:minor` label added.

## Testing
- [x] New tests added (e2e coverage in `admin-volunteer-profile.spec.ts`: badge display for both states, resend success toast, and the full mark-verified flow including the audit note)
- [ ] E2E tests pass (will be verified by CI)

## Checklist

## Additional Notes
- Both endpoints follow the standard admin route pattern (session + `ADMIN` role check, Zod validation, 400 when already verified).
- The user update and AdminNote creation happen in a single transaction.
- `/api/test/users` previously force-set `emailVerified: true`, ignoring overrides; it now respects an explicit `emailVerified: false` so e2e tests can create unverified users.
- The mark-verified e2e test reloads the page before asserting on the admin note because `AdminNotesManager` fetches notes on mount and `router.refresh()` does not remount client components.
- Each mutating e2e test uses a dedicated unverified user since the suite runs fully parallel.
- Local typecheck has pre-existing failures unrelated to this change (missing `@opentelemetry/*` packages in the worktree's node_modules and a stale generated Prisma client for `lastScrapedContent`); the touched files typecheck clean.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -175,8 +175,12 @@ jobs:
           path: web/playwright-report/
           retention-days: 30
 
+      # Best-effort: the HTML report deploy must never fail CI when the tests
+      # themselves passed (e.g. Vercel outages or fair-use blocks). The merged
+      # report is still available as a workflow artifact either way.
       - name: Deploy Playwright report to Vercel
         id: deploy_report
+        continue-on-error: true
         run: |
           cd web/playwright-report
 
@@ -200,7 +204,7 @@ jobs:
           echo "url=$DEPLOY_URL" >> $GITHUB_OUTPUT
 
       - name: Add/Update Playwright Report Comment on PR
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && env.DEPLOY_URL != ''
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_REPO: ${{ github.repository }}
--- a/web/src/app/admin/volunteers/[id]/page.tsx
+++ b/web/src/app/admin/volunteers/[id]/page.tsx
@@ -27,6 +27,8 @@ import {
   Archive,
   ArchiveRestore,
   Megaphone,
+  MailCheck,
+  MailWarning,
 } from "lucide-react";
 import { cn } from "@/lib/utils";
 import { AdminPageWrapper } from "@/components/admin-page-wrapper";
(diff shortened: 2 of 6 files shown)


malinmalliyawadu:
Addressed the review feedback in 50698da2:

- **#2 (note validation)**: added an explicit Zod error message for the 500-char limit.
- **#4 (`.catch(() => ({}))`)**: kept the pattern (it matches the unarchive route, and an empty body is valid since `note` is optional) but added the clarifying comment.
- **#5 (error-case coverage)**: added e2e API assertions that both resend and mark-verified return 400 with "Email is already verified" for a verified volunteer.

Not changed:

- **#1 (test race condition)**: the transaction commits before the PATCH responds, and the reload only happens after the client receives that response, so the note is always committed by then; `toBeVisible()` also auto-waits with the default 5s timeout.
- **#3 (email in toast)**: the confirmation dialog shows the exact email immediately before the action, and `displayName` already falls back to the email when no name is set.

malinmalliyawadu:
CI note: all 20 Playwright shards (including the new email verification tests) and lint passed, but the `merge-reports` job failed with Vercel's "Team exceeded our fair use limits" block when deploying the HTML report, which in turn failed the `Tests passed` gate. That is the known infrastructure issue fixed by #1083, so I cherry-picked that commit (6702edaf) into this branch to unblock the gate - it is content-identical, so it merges cleanly whichever PR lands first.

PR #1107, opened 2026-07-08
fix(mobile): unblock shift signups for volunteers with incomplete profiles

malinmalliyawadu:
# Pull Request

## Description

Volunteers who registered via mobile or OAuth got permanently stuck unable to sign up for shifts: the signup endpoint requires `profileCompleted`, but the mobile app had **no date of birth field**, the mobile profile PUT **didn't accept** the required fields, and it **never recomputed** `profileCompleted` — so the gate could never open from the app. (Agreements were already captured at the post-login gate, which also recomputes the flag; DOB and the missing recompute were the real blockers.)

### Backend (`web/`)

- **`PUT /api/mobile/profile`** now accepts `dateOfBirth` (set-once; only admins can change it after, mirroring the web rule — including `requiresParentalConsent` recompute, under-16 auto-labeling, and admin notifications) plus the two agreement flags, and flips `profileCompleted` via the shared `isProfileComplete` helper with the same welcome email + funnel event as the web `PUT /api/profile`.
- **`GET /api/mobile/profile`** returns the agreement flags and `profileCompleted`, and **self-heals** a stale flag for volunteers who already filled everything in before this fix.
- **`POST /api/mobile/shifts/[id]/signup`** also self-heals a stale flag instead of rejecting, and its 403 now includes `missingFields`.
- **`GET /api/mobile/shifts/[id]`** eligibility now includes `missingProfileFields`, via a new shared `getMissingProfileFields()` in `profile-completion.ts` so labels can't drift between surfaces.

### Mobile (`mobile/`)

- **Profile edit screen**: required Date of Birth field (DD/MM/YYYY masked numeric input, shown locked with a hint once set), required-field markers on phone + emergency contact fields with a "needed before you can sign up for shifts" hint, and an Agreements section reusing the existing `AgreementGate`/`AgreementModal` read-to-agree flow. Saving invalidates shift queries so the signup CTA unlocks immediately.
- **Shift signup sheet**: shows the incomplete-profile warning proactively from the eligibility payload, listing exactly which fields are missing (refreshed from the server 403 body via a new `ApiError.data` field), with the existing one-tap "Complete Profile" button.
- Also fixes 7 pre-existing mobile lint problems (unescaped entities, unused `router`/`FriendStackAvatar`, `Array<T>` style, `useCallback` dependency churn in the feed screen).

## Type of Change

## Version Bump Required
`version:patch`

## Testing
- [x] Unit tests pass (web 449 incl. 13 new for `profile-completion`; mobile 87)
- [x] Manual testing completed — full flow driven via curl against a local dev server:
  - register → signup 403 with `["Mobile number", "Date of birth", "Emergency contact name", "Emergency contact phone"]`
  - mobile PUT with phone/DOB/emergency contact → response `profileCompleted: true` → signup succeeds (PENDING)
  - DOB change rejected 403, same-DOB resend 200, future DOB 400
  - under-16 DOB → consent flag set, admin notification created, "Under 16" label applied
  - stale `profileCompleted=false` with complete fields → GET self-heals

## Checklist
- [x] My changes generate no new warnings (lint + typecheck clean in both apps)

## Additional Notes
- The verify-email flow already sends the "profile completion" welcome email on verification; the completion email from the mobile PUT matches the existing web PUT behavior, so a credentials user can receive both — pre-existing pattern, unchanged.
- The new RN screens passed typecheck/lint/tests and reuse proven components, but haven't been eyeballed on a device/simulator — worth a visual pass before release.

### DIFF
--- a/mobile/app/(auth)/register.tsx
+++ b/mobile/app/(auth)/register.tsx
@@ -269,7 +269,7 @@ export default function RegisterScreen({
               Create your account
             </ThemedText>
             <ThemedText style={[styles.cardSubtitle, { color: mutedText }]}>
-              A few details and you're ready to volunteer
+              A few details and you&apos;re ready to volunteer
             </ThemedText>
 
             {/* Quick sign-up with a provider */}
@@ -498,7 +498,7 @@ export default function RegisterScreen({
                 <ThemedText
                   style={[styles.fieldError, { color: colors.destructive }]}
                 >
-                  Passwords don't match
+                  Passwords don&apos;t match
                 </ThemedText>
               )}
             </View>
--- a/mobile/app/(tabs)/index.tsx
+++ b/mobile/app/(tabs)/index.tsx
@@ -157,21 +157,21 @@ export default function HomeScreen() {
     [apiToggleLike, updateFeedItem]
   );
 
-  const ME_AS_LIKER: LikeUser = {
-    id: profile?.id ?? "",
-    name: "You",
-    profilePhotoUrl: profile?.image ?? undefined,
-  };
-
   const getLikersForItem = useCallback(
     (item: FeedItem): LikeUser[] => {
       const myId = profile?.id;
       const others = myId
         ? item.recentLikers.filter((l) => l.id !== myId)
         : item.recentLikers;
-      return item.likedByMe ? [ME_AS_LIKER, ...others] : others;
+      if (!item.likedByMe) return others;
+      const meAsLiker: LikeUser = {
+        id: myId ?? "",
+        name: "You",
+        profilePhotoUrl: profile?.image ?? undefined,
+      };
+      return [meAsLiker, ...others];
     },
-    [ME_AS_LIKER, profile?.id]
+    [profile?.id, profile?.image]
   );
 
   const handleOpenSheet = useCallback((item: FeedItem) => {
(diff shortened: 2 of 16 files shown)


malinmalliyawadu:
Responding to the automated review findings:

1. **DOB comparison inconsistency** — not an actual mismatch. Client rejects when `date > oneYearAgo`; server accepts when `date <= oneYearAgo`. These are exact logical complements, so the semantics (including the boundary date) are identical on both sides.
2. **Null check in signup self-heal** — the route already returns early on a null user before the self-heal runs; the remaining scenario (user deleted mid-request) is a race that would surface as a 500 via the route's error handler and exists identically across the codebase, so no one-off guard added.
3. **Hard-coded field labels** — the labels live in exactly one place (`getMissingProfileFields`); the mobile app renders the server-provided strings, so there's no duplication to drift.
4. **Un-awaited `invalidateQueries`** — deliberate: the shift screen stays mounted in the nav stack so its query refetches in the background (and the signup sheet re-syncs from props) while back-navigation stays instant. Documented with a comment in 205e8db0.
5. **Agreement checkbox desync** — can't occur: `AgreementGate` is not a toggle. Tapping an accepted agreement reopens the read-only reader ("Agreed — tap to review"); the only mutation is agree → true, so there is no uncheck path to desync.

PR #1108, opened 2026-07-09
fix(mobile): prevent intermittent iOS crash on logout and relaunch

malinmalliyawadu:
## Problem

Sometimes signing out on iOS hard-crashes the app to the home screen, after which the app can fail to re-open (crash loop at launch). No trace in PostHog because the crash is native and kills buffered events.

## Root cause analysis

Both symptoms share one code path: mounting `LoginScreen`.

- On logout, `AuthGate` unmounts the **entire expo-router `<Stack>` navigator** and mounts `LoginScreen` (five reanimated `entering` layout animations) in the same beat, racing the native sign-out alert's dismissal animation. This tree-swap-during-animation race is a known intermittent `EXC_BAD_ACCESS` class on the New Architecture (reanimated [#9402](https://github.com/software-mansion/react-native-reanimated/issues/9402), [#9293](https://github.com/software-mansion/react-native-reanimated/issues/9293), screens [#3648](https://github.com/software-mansion/react-native-screens/issues/3648)).
- The auth token is already deleted by then, so every relaunch cold-starts into the same animation-heavy `LoginScreen` mount, which can crash-loop.

## Changes

- **Defer the auth-state flip** in `logout` / `deleteAccount` until the confirm alert's dismissal has settled (fixed 450ms beat + `InteractionManager`; UIAlertController dismissal isn't tracked by InteractionManager alone)
- **Remove reanimated `entering` animations from `LoginScreen`** - it mounts during the logout tree-swap and on cold start; the hero photo's `expo-image` 600ms fade remains as the screen's only mount motion, layout unchanged
- **Harden `logout`** with try/catch so keychain/network failures can't strand a half-signed-out session (stale server push tokens get pruned via DeviceNotRegistered)
- **Bump `react-native-reanimated` 4.2.1 → 4.2.3** - it was pinned exactly so upstream stability patches never arrived; `expo install --check` confirms compatibility

## Verification

- `tsc --noEmit` clean, `eslint` clean on changed files
- 87/87 mobile unit tests pass
- Needs a device test of the full sign-out → sign-in cycle on iOS (intermittent native crash, not reproducible in CI)


malinmalliyawadu:
Addressed the review in bbc9f919:

- **Medium (error swallowing)**: cleanup failures in `logout` / `deleteAccount` are now logged via `console.warn` and captured to PostHog (`logout_cleanup_failed` / `delete_account_cleanup_failed`) instead of being swallowed silently.
- **Low (magic number)**: the 450ms delay is extracted to `ALERT_DISMISSAL_SETTLE_MS`.
- **HIGH (deleteAccount ordering)**: not applied - the finding doesn't hold. The guard's contract is that the state flip lands at least 450ms after the confirm button press (when the dismissal animation starts). Because `settleUiBeforeTreeSwap()` runs *after* the API call, `set()` always executes >=450ms after the press in every path. On a slow network the alert has long finished dismissing by the time the flip happens, so there is no race left to guard - the settle is redundant there, not useless in a harmful way. `logout` has the same ordering (network first, then settle). Moving the settle before the API call would gain no safety and only delay the actual deletion by 450ms.

Verified: tsc clean, eslint clean, 87/87 unit tests pass.

PR #1142, opened 2026-07-23
fix(locations): stop renamed locations resurrecting as duplicates

malinmalliyawadu:
# Pull Request

## Description

Renaming a location kept producing duplicate locations (one renamed a couple of times now exists three times in prod, with duplicate shifts and shift templates). Root cause analysis and fixes:

1. **`seed-production.ts` resurrection (main culprit).** The seed upserts locations *and ~18 starter shift templates* keyed by the hardcoded original names (`Wellington`, `Glen Innes`, `Onehunga`). After a rename, `where: { name: ... }` matches nothing, so any re-run against a live database re-creates the old name as a brand-new active location plus its templates. Seeding of locations + templates is now **first-boot only** (skipped whenever the `Location` table is non-empty).
2. **Rename cascade gaps.** #1034 added the rename cascade but missed `User.availableLocations` (free-text JSON/legacy list) and `Announcement.targetLocations` (string array). Both feed location-based targeting, so volunteers silently stopped matching after a rename. `availableLocations` is rewritten entry-wise via a new `renameLocationInStoredList()` helper (whole-entry match only - renaming "Wellington" leaves "Wellington City" alone); the string-array rewrites (including the existing `RestaurantManager.locations` one) now dedupe when a row already referenced both names.
3. **Admin merge-locations page.** New `/admin/locations/merge` (linked from the locations settings header) folds a duplicate location into the canonical one. Logic lives in `src/lib/location-merge.ts`, exposed via `POST /api/admin/locations/merge` (dry run by default, `apply: true` to execute). Admins pick the duplicate and the location to keep, get a full preview, and confirm behind an alert dialog. Policy:
   - deletes only *empty* duplicate shifts that have an identical (type/start/end) twin at the target; shifts with signups/walk-ins are never deleted - they're moved and flagged for manual consolidation
   - deletes resurrected duplicate templates (same name already at target), repoints the rest
   - back-fills blank fields on colliding `MealsServed` rows (target row wins); target wins on `DailyMenu`/`MessagingHours` collisions
   - repoints/rewrites all other name references, then deletes the duplicate `Location` row
   - the apply path recomputes the plan inside the transaction, so a stale preview can never act on outdated data
   - orphaned shift venues (shifts referencing a name with no `Location` row) are offered as merge sources too

Not touched (intentionally): `ShortageNotificationLog` (historical log), and the historical importers' auto-create of unknown location names.

## Type of Change

## Version Bump Required
`version:minor`

## Testing
- [x] Unit tests pass (504/504, including new tests for `renameLocationInStoredList` and the shift twin-partitioning logic)
- [x] E2E tests pass (new `admin-merge-locations.spec.ts` covers the full select → preview → confirm → verify flow)

Manual verification against a local database:
- Fabricated the exact prod mess (duplicate location, colliding + unique templates, empty/busy twin shifts, `MealsServed` collision, user/announcement references) and drove the merge through the admin UI: preview counts, twin-shift follow-up warning, confirmation dialog, and success summary all matched; DB state verified row-by-row afterwards (empty twin deleted, busy twin moved with its signup intact, duplicate template removed, `availableLocations` rewritten, old `Location` row gone, zero stray references).
- Renamed a location through the real admin `PATCH /api/admin/locations` on a running dev server: shifts, `availableLocations` JSON, and announcement targets all cascaded.

## Additional Notes

After merging, prod cleanup happens in the admin UI: **Admin → Restaurant Locations → Merge duplicates**, once per stale name (preview first - it lists any duplicate shifts with signups that need manual consolidation afterwards). Also worth checking the marketing CMS `menuLocationName` join key matches the current name.


malinmalliyawadu:
Addressed the review feedback in eb461848:

- **#3 (N+1 in MealsServed collision handling):** fixed - the surviving target rows are now batch-fetched in one query and looked up from a map inside the loop.
- **#5 (orphan venue e2e coverage):** added - a new e2e test creates a shift under a name with no `Location` row, merges it through the UI, and verifies zero remaining references.

Not changed, with reasoning:

- **#1 (SQL injection in `$executeRaw`):** no change needed - Prisma's tagged-template `$executeRaw` is a parameterized query: each `${...}` interpolation is sent as a bind parameter (`$1`, `$2`), never string-concatenated into the SQL. The suggested `$executeRawUnsafe` variant is the API that *can* introduce injection and is discouraged by Prisma's own docs. The tagged-template form used here is the safe one.
- **#4 (seed template config refactor):** deferred as tech debt per the review's own recommendation - the block now only runs on first boot of an empty database, so the maintenance surface is minimal.
- **#6 (CHANGELOG):** it's auto-generated from PR labels; `version:minor` is applied.
- **#7 (filter predicate comment):** the surrounding comment in `computeMerge` already explains the substring over-match and why the helper's null return filters rows out.

PR #1143, opened 2026-07-23
fix: load newly created locations immediately on create-shift page

malinmalliyawadu:
## Description
Newly created locations did not appear across the app until the server was restarted. This fixes the staleness at its root.

**Root cause:** `web/src/lib/locations.ts` built its `LOCATIONS` list and `LOCATION_ADDRESSES` map with a **top-level `await`**. Top-level await runs once, when the module is first imported, and the result is cached for the entire lifetime of the server process - so any location created afterward was invisible until a restart/redeploy.

**Fix:** removed the module-load snapshot entirely. `locations.ts` now exposes fresh async helpers - `getActiveLocationNames()` and `getLocationAddresses()` - that read current data on every call, and every consumer was migrated off the cached constants.

### What changed
- **Location name lists** (dropdowns / filters / selectors) now query fresh: create-shift page, admin dashboard, regulars, restaurant-managers, auto-accept-rules, notifications, all analytics pages (coverage, engagement, milestones, recruitment, shortage-notifications, overview), volunteer detail, and the shifts admin filter.
- **Location validation**: `createLocationEnum()` is now async and built per-request; the two admin `regulars` API routes construct their Zod schemas fresh so a new location validates immediately.
- **Address map**: `calendar-utils` takes an address-map argument (threaded from callers via `getLocationAddresses()`) instead of importing the constant, and `seo`'s `buildShiftEventSchema` takes a caller-resolved address. Email, ICS download, Google Maps links, signup-confirmation, admin assign, and the shift detail pages all pass a fresh map. Unknown locations still fall back to the raw name.
- `Location` is now a plain `string` alias (previously a union derived from the DB snapshot), so runtime-created locations are always type-valid.

## Type of Change

## Version Bump Required
- `version:patch`

## Testing
- [x] `npm run typecheck` - clean (full project)
- [x] `npm run lint` - clean
- [x] `npm run test:run` - 504/504 unit tests pass (includes calendar-utils; the `./locations` test mock was updated to the new async helpers)
- [ ] Manual: create a location in `/admin/locations`, then confirm it appears in the shift-creation forms, admin filters, and analytics selectors without a restart

## Additional Notes
The address-map path touches transactional email + ICS generation, so it was verified via typecheck and the existing unit suite rather than a live email send. `DEFAULT_LOCATION` (a hardcoded `"Wellington"`) and the separate `VOLUNTEER_LOCATIONS` marketing list were already static and are unaffected.

### DIFF
--- a/web/src/app/admin/analytics/coverage/page.tsx
+++ b/web/src/app/admin/analytics/coverage/page.tsx
@@ -4,7 +4,7 @@ import { authOptions } from "@/lib/auth-options";
 import { AdminPageWrapper } from "@/components/admin-page-wrapper";
 import { PageContainer } from "@/components/page-container";
 import { CoverageAnalyticsClient } from "./coverage-analytics-client";
-import { LOCATIONS } from "@/lib/locations";
+import { getActiveLocationNames } from "@/lib/locations";
 import { getShiftCoverage } from "@/lib/shift-coverage";
 import { parseDaysParam } from "@/lib/parse-days-param";
 
@@ -32,7 +32,7 @@ export default async function CoverageAnalyticsPage({
 
   const data = await getShiftCoverage(parseInt(months, 10), location, daysFilter);
 
-  const locationOptions = LOCATIONS.map((loc) => ({
+  const locationOptions = (await getActiveLocationNames()).map((loc) => ({
     value: loc,
     label: loc,
   }));
--- a/web/src/app/admin/analytics/engagement/page.tsx
+++ b/web/src/app/admin/analytics/engagement/page.tsx
@@ -4,7 +4,7 @@ import { authOptions } from "@/lib/auth-options";
 import { AdminPageWrapper } from "@/components/admin-page-wrapper";
 import { PageContainer } from "@/components/page-container";
 import { EngagementAnalyticsClient } from "./engagement-analytics-client";
-import { LOCATIONS } from "@/lib/locations";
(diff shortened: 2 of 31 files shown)


malinmalliyawadu:
Thanks for the thorough review! Addressed below.

**3. Test coverage — done.** Added `web/src/lib/locations.test.ts` covering `getActiveLocationNames()` and `getLocationAddresses()`, plus an assertion that each call re-queries (no module-scoped memoization) so we can't silently regress back to the cached-snapshot behaviour. Full suite is 507 green.

**2. Async validation schema (flagged High) — no change needed.** This isn't actually the anti-pattern it looks like. `createLocationEnum()` is called exactly **once per request** in each of the two `regulars` routes (`buildCreateRegularVolunteerSchema(await createLocationEnum())` / `buildUpdateRegularVolunteerSchema(...)`), so it's one query per request, not per-item and never in a loop — which is precisely the "fetch once, then build the schema" pattern the suggestion recommends, just packaged as a small factory. The `createShift`/`createTemplate` server actions don't use the enum at all (`location: z.string().min(1)`), so there's no hidden per-item query there either.

**1. Short-TTL cache — intentionally not doing this.** A 5–10s in-memory cache would re-introduce exactly the staleness this PR removes (a location created inside the window wouldn't appear until it expired), which defeats the fix. As the review notes, the query hits a tiny indexed table at hundreds-of-requests/day, so the cost is negligible and immediacy is the whole point. Happy to revisit if load ever changes.

PR #1144, opened 2026-07-23
fix(admin): render 404s inside the admin dashboard chrome

malinmalliyawadu:
## Description

Admin 404 pages were rendering as Next.js's bare default 404 - no sidebar, no header - leaving admins with no easy way back to a working page.

**Root cause:** there was no `not-found.tsx` anywhere in the app, and Next.js only renders a *nested* `not-found.tsx` for explicit `notFound()` calls. Unmatched URLs (e.g. `/admin/typo`) fall straight through to the chrome-less root 404.

**Fix - two files:**

- `web/src/app/admin/not-found.tsx` - the 404 boundary. Because it lives under `admin/layout.tsx`, it renders inside the full admin chrome (sidebar + header). Uses `AdminPageWrapper` to set the header title and offers two escape routes: **Back to dashboard** (`/admin`) and **Go back** (`router.back()`), styled to match the existing admin design language.
- `web/src/app/admin/[...not-found]/page.tsx` - a lowest-priority catch-all that calls `notFound()`, so unmatched `/admin/*` URLs render the boundary above *with chrome* instead of the bare root 404. Real admin routes always take precedence over the catch-all.

## Type of Change

## Version Bump Required
`version:patch`

## Testing
- [x] `npm run typecheck` passes
- [x] `eslint` clean on both new files
- [x] `npm run build` - `✓ Compiled successfully`; Next's route collection accepted the catch-all + not-found with no route conflicts (the build's non-zero exit is only the post-compile PostHog source-map upload failing on a missing local API key, unrelated to this change and green in CI)

## Additional Notes

Scope is the **admin** area, as requested. Non-admin 404s still hit Next's default page - a follow-up could add a root `src/app/not-found.tsx` with the public site header/footer for consistent site-wide 404 chrome.

No live browser check: it requires an admin login, and the repo-root `launch.json`'s `web` config points at the main checkout rather than this worktree, so it wouldn't exercise these files. The production build directly validates the routing behavior that was the actual risk.

### DIFF
--- a/web/src/app/admin/[...not-found]/page.tsx
+++ b/web/src/app/admin/[...not-found]/page.tsx
@@ -1,0 +1,14 @@
+import { notFound } from "next/navigation";
+
+/**
+ * Catch-all for unmatched `/admin/*` URLs.
+ *
+ * Next.js only renders a nested `not-found.tsx` for explicit `notFound()`
+ * calls — unmatched URLs fall back to the root (chrome-less) 404. This
+ * lowest-priority catch-all intercepts those URLs and calls `notFound()`, so
+ * they render `admin/not-found.tsx` inside the admin layout (sidebar + header)
+ * instead. Real admin routes always take precedence over this catch-all.
+ */
+export default function AdminCatchAllNotFound() {
+  notFound();
+}
--- a/web/src/app/admin/not-found.tsx
+++ b/web/src/app/admin/not-found.tsx
@@ -1,0 +1,63 @@
+"use client";
+
+import Link from "next/link";
+import { useRouter } from "next/navigation";
+import { ArrowLeft, Compass, LayoutDashboard } from "lucide-react";
+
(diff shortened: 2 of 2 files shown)


malinmalliyawadu:
This PR was squash-merged before the review fixes could land, so I applied them in a follow-up: #1146.

Addressed there:
- **Motion animation** — content wrapped in `MotionPageContainer` (CLAUDE.md motion.dev requirement).
- **`router.back()` edge case** — added a fallback to `/admin` when there's no in-app history.
- **Test coverage** — new `admin-404.spec.ts` (3 tests, all passing): unmatched URL renders with admin chrome, deeply-nested unmatched URL, and dashboard-link recovery.
- **Nit** — fixed an em dash in the body copy.

Kept the `Compass` icon and the "wrong turn" copy as-is (brand voice, playful-but-clear).

Issue #1125, opened 2026-07-17
Notifications for non default locations not working


malinmalliyawadu:
The push notifications for different locations than the one you are currently on don't work

PR #1145, opened 2026-07-23
fix(admin): include selected locations in users page location filter

malinmalliyawadu:
# Pull Request

## Description
The location filter on the admin All Users page (`/admin/users`) only matched a user's `defaultLocation`, so anyone whose filtered location was only in their selected locations (`availableLocations`) never appeared. This notably hid admins who work across locations without a matching default.

The filter now matches a user's default location OR their selected locations by default, and a new "Default location only" checkbox next to the location dropdown restores the strict behavior. The choice is encoded as a `locationScope=default` URL param that survives search, role/archived filters, sorting, pagination and link sharing.

Details:
- Both query paths are covered: the Prisma where clause and the raw SQL path used when sorting by shifts.
- `availableLocations` matching handles both storage formats (JSON-stringified array and legacy plain text), following the same precedent as the announcement-targeting fix for #1125.
- The checkbox only renders while a location filter is active.

## Type of Change
- [ ] 📚 Documentation update
- [ ] 🔧 Build/CI changes

## Version Bump Required
`version:patch`

## Testing
- [x] Unit tests pass
- [x] E2E tests pass

New e2e test in `admin-users.spec.ts` seeds one user with Glen Innes as their default location and one with it only in their selected locations, then verifies the broad match shows both, the toggle narrows to the default-only user, and unchecking widens again.

Manually verified against the demo seed data: filtering Glen Innes showed all 52 associated users (including an admin whose default was Wellington), toggling narrowed to the single user whose default is Glen Innes, and both behaviors also hold on the shifts-sorted (raw SQL) path.

## Checklist
- [ ] Any dependent changes have been merged and published

## Additional Notes
Behavior change to be aware of: the default meaning of the location filter is now "anyone associated with this location" rather than "default location only". The old behavior remains one click away via the new checkbox.

### DIFF
--- a/web/src/app/admin/users/page.tsx
+++ b/web/src/app/admin/users/page.tsx
@@ -40,6 +40,13 @@ export default async function AdminUsersPage({
   const locationFilter = Array.isArray(params.location)
     ? params.location[0]
     : params.location;
+  const locationScopeRaw = Array.isArray(params.locationScope)
+    ? params.locationScope[0]
+    : params.locationScope;
+  // "all" matches a user's default location OR any of their selected
+  // (available) locations; "default" restricts to the default location only.
+  const locationScope: "all" | "default" =
+    locationScopeRaw === "default" ? "default" : "all";
   const archivedFilterRaw = Array.isArray(params.archived)
     ? params.archived[0]
     : params.archived;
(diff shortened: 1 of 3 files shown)


malinmalliyawadu:
Thanks for the review. Addressed in 770a8846:

**Done**
- Added the legacy plain-text `availableLocations` case to the e2e test (third seeded user with `availableLocations: "Glen Innes"`).
- The same test now repeats the broad and default-only assertions with `sortBy=signups`, so the raw SQL query path is exercised end to end.

**Declined, with reasons**
- **`::jsonb` operators / quote escaping**: casting `availableLocations::jsonb` is a known trap in this codebase - the column still holds legacy non-JSON values ("Wellington", "Wellington, Glen Innes"), and PostgreSQL evaluates the cast per row, so it throws `invalid input syntax for type json` regardless of OR ordering. That is exactly why `safeParseLocations` exists and why the three existing match sites (`recruitment-users.ts`, `milestone-users.ts`, `announcement-targeting.ts`) use string matching; this PR follows that precedent. There is no injection risk since the value is parameterized. Quote escaping guards an impossible input: location names come from the admin-managed `Location` table and the filter dropdown, and the rename flow (#1142) keeps stored lists in sync with those canonical names.
- **Case-insensitive matching**: same reasoning - values on both sides come from the same canonical `Location.name`, and the existing match sites are case-sensitive. Making only this one insensitive would create inconsistency without fixing a real scenario.
- **Mobile label split**: verified in a narrow viewport; "Default location only" fits comfortably in the stacked mobile layout.
- **GIN index / jsonb migration**: agreed at scale, but out of scope for this fix and only worth doing as a proper migration of the column across all call sites.

PR #1154, opened 2026-07-24
feat(announcements): target volunteers by shift history

malinmalliyawadu:
# Pull Request

## Description

Admins could only reach volunteers by **profile location**, which matches everyone who lists a restaurant - including people who signed up two years ago and never turned up. To reach "volunteers who worked at Onehunga in the last 3 months" the only option was picking each person by hand in the Specific Volunteers search; there was no import, and the Specific Shifts picker only lists upcoming shifts, so past shifts couldn't be selected either.

This adds a **Shift History** targeting dimension to `/admin/announcements`:

> worked at `[locations]` · shift finished between `[dates]` · at least `[N]` shifts

The audience is derived from data the portal already holds, so no list has to be assembled, uploaded, or kept up to date, and the same filter can be reused for later sends.

The difference is not cosmetic. On production-shaped data, "worked at Onehunga in the last 3 months" is **35 volunteers**; "profile lists Onehunga" is **23**. The Locations block now says what it actually matches and points at Shift History for people who turned up.

## Type of Change

## Version Bump Required

`version:minor` (label applied)

## How it works

**Definition of a worked shift.** A `CONFIRMED` signup on a shift whose `end` has passed. Waitlisted and no-show signups are deliberately excluded - a no-show never worked the shift, so they don't belong in a "thanks for volunteering" audience. This is narrower than `ANNOUNCEMENT_SHIFT_TARGET_STATUSES`, which the existing Specific Shifts dimension keeps using unchanged.

**Dormant by default.** The dimension is off unless `targetActivityMinShifts` is set. Every existing announcement keeps its exact recipient set, and the recipient counts shown on the announcements list are unchanged.

**Dates.** The window applies to when the shift *finished*, and `YYYY-MM-DD` inputs are anchored to NZ calendar days - "from 24 April" is midnight 24 April NZ, "to 24 April" is the end of that day - so a range reads inclusively the way an admin picked it. DST is handled via `createNZDate`; both sides of the transition are covered by tests.

**Consistency between channels.** `web/src/app/api/mobile/feed/route.ts` filters announcements for one known user in app code rather than SQL, so it gets an in-memory twin of the SQL condition (`userMatchesActivityTargeting`), living next to it in the same file with a comment tying them together. Without this, a volunteer could be emailed an announcement that never appeared in their feed.

**One parser, two routes.** The create route and the recipient-count route previously each hand-rolled their own body parsing. They now share `parseTargetingFromRequest`, so the live count an admin sees can't drift from the audience that actually gets sent to - the exact class of bug a new dimension would otherwise invite.

## Testing

- [x] Unit tests pass - 520 passing, 13 new in `announcement-targeting.test.ts` covering NZ date anchoring (both DST sides), malformed-date handling, min-shift clamping, and every branch of the matcher
- [x] New tests added
- [ ] E2E tests pass - not run locally; no existing announcements spec

Verified live against a dev database with realistic seed data:

| Targeting | Recipients |
|---|---|
| No targeting | 87 |
| Onehunga, last 3 months, min 1 | **35** |
| Onehunga, last 3 months, min 3 | 2 |
| Any location, last 3 months | 75 |
| Onehunga, all time | 41 |
| Onehunga activity AND Wellington profile | 30 |
| Window starting in the future | 0 |

The 35 matches a direct SQL check against the same data. Published a test announcement end to end (email and push off): stored `targetActivityFrom` was `2026-04-23 12:00 UTC` - correctly midnight 24 April NZST - and the saved announcement's recipient count and summary line ("worked 1+ shift at Onehunga since 24 Apr 2026") both rendered correctly. Test row deleted afterwards.

Checked at 1440px, 420px, and in dark mode. One defect found and fixed during review: the selected date preset rendered in a near-white `secondary` that was indistinguishable from the unselected outline buttons, so selection state was effectively invisible. Active presets are now primary green with `aria-pressed`.

## Checklist

- [x] I have added tests that prove my feature works

## Additional Notes

**Migration.** `20260724000000_add_announcement_activity_targeting` adds four nullable/defaulted columns to `Announcement`. Additive and backward compatible; needs `prisma:deploy` on release. No new indexes - the existing `Signup(userId, canceledAt)` and `Shift(end)` indexes serve the recipient join.

**Out of scope, worth a look separately.** The recipient query has no `archivedAt` filter, so announcements to "all volunteers" also email and push to archived accounts (archived for `INACTIVE_12_MONTHS` or `NEVER_ACTIVATED`). Every other admin surface filters those out by default. That is pre-existing behaviour and changing it would shift the counts displayed for existing announcements, so I left it alone rather than folding it into this PR.

### DIFF
--- a/web/prisma/migrations/20260724000000_add_announcement_activity_targeting/migration.sql
+++ b/web/prisma/migrations/20260724000000_add_announcement_activity_targeting/migration.sql
@@ -1,0 +1,14 @@
+-- Shift-history targeting for announcements: reach the volunteers who actually
+-- worked shifts matching a location / date-window / minimum-count filter,
+-- rather than everyone whose profile merely lists a location.
+--
+-- The dimension is dormant unless "targetActivityMinShifts" is set, so every
+-- existing announcement keeps its current recipient set.
+--
+-- The recipient query joins Signup to Shift; the existing
+-- Signup(userId, canceledAt) and Shift(end) indexes already serve it.
+ALTER TABLE "Announcement"
+  ADD COLUMN "targetActivityLocations" TEXT[] DEFAULT ARRAY[]::TEXT[],
+  ADD COLUMN "targetActivityFrom" TIMESTAMP(3),
+  ADD COLUMN "targetActivityTo" TIMESTAMP(3),
+  ADD COLUMN "targetActivityMinShifts" INTEGER;
--- a/web/prisma/schema.prisma
+++ b/web/prisma/schema.prisma
@@ -1023,6 +1023,14 @@ model Announcement {
   targetUserIds   String[] @default([]) // Specific User IDs
   targetShiftIds  String[] @default([]) // Users with non-cancelled signups on these shifts
 
+  // Shift-history targeting — volunteers who actually worked shifts matching
+  // these criteria. Dormant unless targetActivityMinShifts is set; the other
+  // three fields narrow the window (all optional, empty/null = no narrowing).
+  targetActivityLocations String[]  @default([]) // Shift locations worked at
+  targetActivityFrom      DateTime? // Shift must have ended on/after this
+  targetActivityTo        DateTime? // Shift must have ended on/before this
+  targetActivityMinShifts Int? // Minimum matching shifts; null = dimension off
+
   // Email
   sendEmail   Boolean   @default(false)
   emailSentAt DateTime?
--- a/web/src/app/admin/announcements/announcements-content.tsx
+++ b/web/src/app/admin/announcements/announcements-content.tsx
@@ -3,7 +3,7 @@
 import { useState, useRef, useCallback, useEffect, useMemo } from "react";
 import { usePathname, useRouter, useSearchParams } from "next/navigation";
 import ReactMarkdown from "react-markdown";
-import { format } from "date-fns";
+import { format, subMonths } from "date-fns";
 import {
   Plus,
   Trash2,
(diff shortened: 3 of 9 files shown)


malinmalliyawadu:
Thanks — addressed in #1156, since this PR merged before the follow-up could be pushed to it.

**Invalid date logging (#4)** — this turned out to be a real bug rather than a missing log line. `Number.isNaN(date.getTime())` never fires for a well-formed but impossible date, because the constructor rolls the components over: `2026-13-99` becomes a valid date in 2027, and `2026-02-31` becomes 3 March. The guard was dead code, and the bad value silently became a *wrong* window bound — on an email blast, the wrong audience with nothing to show for it. The parsed result is now checked back against the string it came from, and rejections are logged. A test covers `2028-02-29` to prove the check didn't over-tighten onto real leap days.

**Performance (#1)** — the subquery now uses `EXISTS` when the minimum is one shift, so Postgres stops at the first matching row instead of counting every shift the volunteer ever worked. Counts above one still use `COUNT(DISTINCT …)`. Re-ran all seven targeting combinations against dev data: identical counts, so the rewrite is behaviour-preserving.

**Null location comment (#3)** — added, spelling out that `NULL = ANY(array)` yields `NULL` rather than true.

**Composite `Shift(location, end)` index (#2)** — not doing this one. The subquery reaches `Shift` through `Signup`'s foreign key, so rows are fetched by primary key and that index would never be consulted; the driving access path is `Signup(userId, …)`, already indexed. It would cost write throughput for no read benefit.

Also skipped the suggested SQL-vs-in-memory parity test — it needs a live database, and the unit suite mocks Prisma. The two implementations are kept in sync by comment instead, which is weaker; worth revisiting if that file grows.

PR #1155, opened 2026-07-24
fix(announcements): exclude archived volunteers from targeting

malinmalliyawadu:
## Problem

`buildRecipientConditions` in [announcement-targeting.ts](web/src/lib/announcement-targeting.ts) selected from `"User"` with no filter on `archivedAt`. Both callers - `countAnnouncementRecipients` (the admin form's live counter and the per-announcement counts on /admin/announcements) and `findAnnouncementRecipients` (the email and push dispatch paths) - therefore included archived accounts in every targeting dimension except explicitly-named user IDs.

So an announcement sent to "all volunteers", or to a location / grade / label / shift, also emailed and push-notified people the org has deliberately stopped engaging: `INACTIVE_12_MONTHS`, `NEVER_ACTIVATED`, `NEVER_MIGRATED`, `MANUAL`. Archived users also can't sign in (blocked in `auth-options`, `mobile-auth`, and `passkey-auth`), so the announcement email pointed them at a `/dashboard` they can't reach.

This looks like an oversight rather than a deliberate choice - every other admin surface that lists volunteers already filters archived accounts out by default:

- `web/src/app/api/admin/users/route.ts` - `{ archivedAt: null }`
- `web/src/app/admin/users/page.tsx` - archived filter defaults to "active"
- `web/src/app/api/admin/volunteers/route.ts`, `.../volunteers/search`, `.../notifications/send-shortage`, the leaderboard, achievements, friends

## Fix

Add an archive condition to `buildRecipientConditions`, with an exemption for explicitly-targeted user IDs:

- No `targetUserIds` -> `"archivedAt" IS NULL`
- With `targetUserIds` -> `("archivedAt" IS NULL OR "User".id = ANY(<ids>))`

The exemption is deliberate: the **Announce** button on an archived volunteer's profile (`/admin/volunteers/[id]`) deep-links to `/admin/announcements?userIds=<id>`, and `/api/admin/users?ids=` hydrates archived users for the badge - so hand-picking an archived volunteer is a supported flow and should keep working. Cross-dimension AND semantics are unchanged: `userIds + grade` still intersects.

## ⚠️ Recipient counts change

`/admin/announcements` recomputes `recipientCount` for **existing** announcements on every load, so counts shown for already-sent announcements will drop by however many archived volunteers matched their targeting. The number of people actually emailed at send time didn't change retroactively - only the displayed figure. Production currently has archived users, so this will be visible there (the dev DB has none, hence the seeding below).

## Verification

Seeded 2 archived + 1 active volunteer into the dev DB (87 users -> 90, all in Wellington/GREEN) and exercised the running app as an admin:

| Targeting | Recipients | Expected |
|---|---|---|
| All volunteers | 88 | 88 active (of 90 total) ✅ |
| Location = Wellington | 79 | archived seeds excluded ✅ |
| Grade = GREEN | 36 | archived seed excluded ✅ |
| `targetUserIds` = 2 archived | 2 | both still reached ✅ |
| `targetUserIds` = 1 archived + 1 active | 2 | ✅ |
| `targetUserIds` + grade | 2 | AND semantics intact ✅ |

The admin form's live counter read **"~88 volunteers"** with 90 users in the DB. Seed rows were deleted afterwards (dev DB back to 87 users, 0 archived).

Also added `web/src/lib/announcement-targeting.test.ts` - unit tests asserting the generated SQL carries `"archivedAt" IS NULL` for broad targeting and the OR-exemption when user IDs are named.

`npm run lint`, `npm run typecheck`, and `npm run test:run` (510 tests) all pass.

### DIFF
--- a/web/src/lib/announcement-targeting.test.ts
+++ b/web/src/lib/announcement-targeting.test.ts
@@ -44,7 +44,9 @@ beforeEach(() => {
 
 describe("announcement recipient targeting", () => {
   it("excludes archived volunteers when targeting everyone", async () => {
-    await countAnnouncementRecipients(emptyTargeting);
+    queryRaw.mockResolvedValueOnce([{ count: BigInt(42) }]);
+    const count = await countAnnouncementRecipients(emptyTargeting);
+    expect(count).toBe(42);
     expect(lastSql()).toContain(`"archivedAt" IS NULL`);
   });
 
@@ -70,5 +72,19 @@ describe("announcement recipient targeting", () => {
     expect(sql).toContain(`( "archivedAt" IS NULL OR "User".id = ANY(`);
     // The exemption reuses the same ids as the targeting condition itself.
     expect(lastValues()).toEqual(["user-1", "user-2", "user-1", "user-2"]);
+  });
+
+  it("keeps cross-dimension AND semantics when ids are combined with a filter", async () => {
+    await findAnnouncementRecipients({
+      ...emptyTargeting,
+      targetUserIds: ["user-1"],
+      targetGrades: ["GREEN"],
+    });
+    const sql = lastSql().replace(/\s+/g, " ");
+    // The archive exemption widens who the id list may reach — it never
+    // relaxes the other dimensions, so a named volunteer of the wrong grade
+    // is still filtered out.
+    expect(sql).toContain(`( "archivedAt" IS NULL OR "User".id = ANY(`);
+    expect(sql).toContain(`"volunteerGrade"::text = ANY(`);
   });
 });
--- a/web/src/lib/announcement-targeting.test.ts
+++ b/web/src/lib/announcement-targeting.test.ts
@@ -1,0 +1,74 @@
+import { vi, describe, it, expect, beforeEach } from "vitest";
+
+vi.mock("@/lib/prisma", () => ({
+  prisma: {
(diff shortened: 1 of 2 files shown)


malinmalliyawadu:
Thanks — addressing the two open points.

**Mixed targeting semantics: intended.** The exemption widens *who the id list may reach*; it never relaxes the other dimensions. `targetUserIds: [archived-green] + targetGrades: ["BLUE"]` still matches nobody, because the grade condition is ANDed independently. Scoping the exemption to "user IDs are the sole dimension" would break the flow it exists for: the **Announce** button on an archived volunteer's profile prefills `?userIds=<id>`, and an admin who then also ticks a location or grade would silently lose the volunteer they explicitly picked.

Added a test covering exactly that case (`keeps cross-dimension AND semantics when ids are combined with a filter`), plus the `countAnnouncementRecipients` return-value assertion you suggested — 43ce80a4.

**Integration coverage** is already done, just not as a committed test: I seeded 2 archived + 1 active volunteer into a real DB and drove the running app as an admin. Broad targeting returned 88 of 90, Wellington 79, GREEN 36, and explicit archived IDs returned 2/2 — the table in the PR description. There's no existing `/admin/announcements` Playwright spec, so a new e2e suite felt like separate scope from this fix.

Skipping the partial index — 87 users in dev, and the sequential scan is nowhere near a bottleneck at this table size.

PR #1156, opened 2026-07-24
fix(announcements): reject impossible activity dates in shift-history targeting

malinmalliyawadu:
# Pull Request

## Description

Follow-up to #1154, which merged before this could be added to it.

The automated review on #1154 suggested logging invalid activity dates. Checking that path turned up a real bug rather than just a missing log line.

`parseActivityDate` guarded with `Number.isNaN(date.getTime())`, but a well-formed yet impossible date never produces an invalid Date - the date constructor rolls the components over into a real date in a later month:

```
"2026-13-99" → a valid date in 2027
"2026-02-31" → 3 March 2026
```

So the guard never fired, and the bad value became a silently *wrong* window bound. On a targeted email blast that means reaching an audience nobody asked for, with nothing in the UI or logs to show it happened. The regex only proves the shape `\d{4}-\d{2}-\d{2}`, not that the date exists.

The parsed result is now checked back against the string it came from (`formatInNZT(date, "yyyy-MM-dd") === value`), and a rejected date is logged instead of disappearing quietly. Rejected bounds still widen the window rather than throwing mid-send, which is the existing deliberate behaviour.

The date inputs on the admin form can't produce these values; a direct API call can.

## Type of Change

## Version Bump Required

`version:patch` (label applied)

## Also in this PR

**EXISTS fast path.** The recipient subquery now uses `EXISTS` when the minimum is one shift - by far the common case - so Postgres stops at the first matching row instead of counting every shift a volunteer ever worked. Counts above one still use `COUNT(DISTINCT …)`. This addresses the performance note in the #1154 review.

**Clarifying comment** on why `NULL` shift locations drop out of the SQL without an explicit `IS NOT NULL` (`NULL = ANY(array)` yields `NULL`, not true) - subtle enough to be worth stating, as the review pointed out.

## Not doing

The review also suggested a composite `Shift(location, end)` index. Declining it: the subquery reaches `Shift` through `Signup`'s foreign key, so rows are looked up by primary key and that index would never be consulted. The driving access path is `Signup(userId, …)`, which is already indexed. Adding it would cost write throughput and disk for no read benefit.

## Testing

- [x] Unit tests pass - 522 total, 2 new
- [x] New tests added

New tests cover both impossible dates and, importantly, a real leap day (`2028-02-29`) to prove the round-trip check didn't over-tighten into rejecting valid dates.

Re-ran the full targeting matrix against dev data after the EXISTS rewrite. Every count is identical to before, confirming it is behaviour-preserving:

| Targeting | Before | After |
|---|---|---|
| No targeting | 87 | 87 |
| Onehunga, last 3 months, min 1 | 35 | 35 |
| Onehunga, last 3 months, min 3 | 2 | 2 |
| Any location, last 3 months | 75 | 75 |
| Onehunga, all time | 41 | 41 |
| Onehunga activity AND Wellington profile | 30 | 30 |
| Window starting in the future | 0 | 0 |

Confirmed the invalid-date path end to end: posting `2026-13-99` returns 41 (the bound correctly dropped, matching "Onehunga, all time") and the server logs `[announcement-targeting] ignoring unparseable activity start date: 2026-13-99`.

## Checklist

- [x] I have added tests that prove my fix is effective

## Additional Notes

No schema or migration changes. Pure logic fix on top of the feature merged in #1154.

### DIFF
--- a/web/src/lib/announcement-targeting.test.ts
+++ b/web/src/lib/announcement-targeting.test.ts
@@ -84,6 +84,32 @@ describe("announcement-targeting", () => {
       expect(t.targetActivityFrom).not.toBeNull();
     });
 
+    it("rejects a leap day in a non-leap year", () => {
+      const t = parseTargetingFromRequest({
+        targetActivityFrom: "2027-02-29",
+        targetActivityMinShifts: 1,
+      });
+
+      expect(t.targetActivityFrom).toBeNull();
+    });
+
+    it("rejects a zero month or day", () => {
+      // The shape regex allows these; they roll backwards into the previous
+      // year rather than failing, so only the round-trip check catches them.
+      expect(
+        parseTargetingFromRequest({
+          targetActivityFrom: "2026-00-01",
+          targetActivityMinShifts: 1,
+        }).targetActivityFrom
+      ).toBeNull();
+      expect(
+        parseTargetingFromRequest({
+          targetActivityFrom: "2026-01-00",
+          targetActivityMinShifts: 1,
+        }).targetActivityFrom
+      ).toBeNull();
+    });
+
     it("clamps the minimum shift count into a sane range", () => {
       expect(
         parseTargetingFromRequest({ targetActivityMinShifts: 0 })
--- a/web/src/lib/announcement-targeting.test.ts
+++ b/web/src/lib/announcement-targeting.test.ts
@@ -62,6 +62,28 @@ describe("announcement-targeting", () => {
       expect(t.targetActivityTo).toBeNull();
     });
 
+    it("rejects well-formed but impossible dates instead of rolling them over", () => {
+      // The date constructor happily turns these into real dates in a later
+      // month. Accepting that would silently shift a window bound.
+      const t = parseTargetingFromRequest({
+        targetActivityFrom: "2026-13-99",
+        targetActivityTo: "2026-02-31",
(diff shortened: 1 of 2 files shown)


malinmalliyawadu:
Thanks — took the two test suggestions, declined the rest. Pushed in 0136a2e.

**Non-leap-year leap day (#2) and zero month/day (#3)** — added both. Your read was right that the round-trip check already catches them, so this is test-only, no production change: `2027-02-29`, `2026-00-01` and `2026-01-00` all roll over rather than failing, and only comparing the result back to the input rejects them. Worth pinning, since the behaviour was undocumented and would quietly regress if that check were ever simplified.

**Tightening the regex (#3)** — not doing it. It would overlap the round-trip check, giving two places that have to agree about which dates are real. The round-trip check is the stricter of the two and can't drift from the calendar, so it stays the single source of truth.

**Structured logging (#1)** — leaving `console.warn`. `console.*` is the house style across `src/lib` (`achievement-calculator.ts`, and the announcements route's own `[announcements] email dispatch failed`), so switching this one call would make it the odd one out. Worth doing app-wide, not here.

**Enum for `edge` (#4)** — `"start" | "end"` is already a literal union, which gives the same exhaustiveness checking as an enum without the runtime object. Leaving as is.

**`roundTrips` naming** — leaving it; it names what the check does rather than its verdict, which is the part worth reading at the call site.

PR #1157, opened 2026-07-24
feat(announcements): redesign admin console with live feed preview

malinmalliyawadu:
# Pull Request

## Description

Complete visual and information-architecture overhaul of `/admin/announcements`. The old page was a single 1,726-line scrolling form that buried the two questions carrying the most operational risk: **what will volunteers actually see**, and **who exactly receives this**. Both answers only appeared after publishing.

The page is now a two-view console split across focused files:

**Overview** — a pulse strip (live in the feed / next to expire / published all-time) over a ledger split into "In the feed now" and "Expired". Each row is a date tile plus title, one-line preview, audience summary and delivery status. Expanding a row asks the API how many volunteers that stored audience matches *today*.

**Composer** — inputs on the left (Message, Schedule, Audience, Delivery), with a sticky rail on the right holding the consequences:

- A **phone silhouette rendering the volunteer's real feed card**, mirroring `mobile/app/(tabs)/index.tsx` down to the 📢 emoji tile and yellow chip, updating as you type.
- The live reach count, with the audience spelled out as plain-English conditions joined by "and" ("based in Wellington" / "and worked exactly 1 shift").
- A publish button labelled with the number it will actually send to — "Publish to ~79 volunteers".

**Targeting** moves from a wall of checkboxes to six collapsible filter rows, each showing its current state in the collapsed header, so the whole audience reads at a glance without opening anything.

Two capabilities were added to targeting along the way:

- **`targetActivityMaxShifts`** bounds shift-history targeting from above, so an announcement can reach volunteers who have worked *exactly one* shift. Threaded through the schema, the SQL condition (`BETWEEN` when a max is set), the in-memory mobile-feed twin, and request parsing — which clamps a crossed range up to the minimum rather than silently matching nobody.
- **`POST /api/admin/announcements/recipients`** returns the matched volunteers using the same `findAnnouncementRecipients` the send path uses (alphabetised, capped at 300). It powers a "see exactly who" toggle under the reach count, with each name linking to their admin profile.

## Type of Change

## Version Bump Required
`version:minor` — new targeting capability and a new API route, backward compatible.

## Testing
- [x] Unit tests pass — 524 passed (50 files), including 6 new cases covering the max-shifts bound and its clamping
- [x] New tests added
- [ ] E2E tests pass — not run locally; no announcements e2e spec exists in the suite

Manual verification in the browser (light, dark, and 375px mobile):

- Published an announcement end-to-end and confirmed it lands in the ledger
- Confirmed the "see exactly who" list stays in step with the count — 87 names with no filters, dropping to 79 in lockstep when shift-history targeting was enabled
- Exercised delete-with-confirmation, and the expired-section split
- Confirmed existing deep links (`?userIds=`, `?shiftIds=`, `?locations=`, `?grades=`, `?labels=`, `?sendEmail=`) still seed and open the composer with real volunteer names hydrated

`npm run lint` and `tsc --noEmit` are both clean.

## Checklist
- [x] I have added tests that prove my feature works

## Additional Notes

**Migration**: `20260724144327_add_announcement_activity_max_shifts` adds one nullable column. No backfill needed — `NULL` means "no upper bound", which is the existing behaviour.

**Backward compatibility**: every existing announcement keeps rendering and matching identically; the new bound is dormant unless an admin sets it.

**Unrelated finding worth flagging** — `web/.env` in local dev has the Supabase variables set to *empty strings*, which look set when you grep. `src/lib/supabase.ts` throws at module load, so every route importing `@/lib/storage` returns 500 in dev: announcement delete, announcement image upload, and `/api/profile/photo`. Not a code bug and prod is unaffected, but it makes those routes look broken locally. Patching the local `.env` with dummy values fixes it.

### DIFF
--- a/web/prisma/migrations/20260724144327_add_announcement_activity_max_shifts/migration.sql
+++ b/web/prisma/migrations/20260724144327_add_announcement_activity_max_shifts/migration.sql
@@ -1,0 +1,3 @@
+-- Add an optional upper bound to shift-history targeting so announcements can
+-- target e.g. volunteers who have worked exactly one shift.
+ALTER TABLE "Announcement" ADD COLUMN "targetActivityMaxShifts" INTEGER;
--- a/web/prisma/schema.prisma
+++ b/web/prisma/schema.prisma
@@ -1030,6 +1030,7 @@ model Announcement {
   targetActivityFrom      DateTime? // Shift must have ended on/after this
   targetActivityTo        DateTime? // Shift must have ended on/before this
   targetActivityMinShifts Int? // Minimum matching shifts; null = dimension off
+  targetActivityMaxShifts Int? // Maximum matching shifts; null = no upper bound
 
   // Email
   sendEmail   Boolean   @default(false)
--- a/web/src/app/admin/announcements/announcement-list.tsx
+++ b/web/src/app/admin/announcements/announcement-list.tsx
@@ -1,0 +1,471 @@
+"use client";
+
+import { useMemo, useState } from "react";
+import ReactMarkdown from "react-markdown";
+import { format, formatDistanceToNow } from "date-fns";
+import {
+  Bell,
+  ChevronDown,
+  Clock,
+  Mail,
+  Megaphone,
+  Search,
+  Trash2,
+  Users,
+} from "lucide-react";
+import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
+import { cn } from "@/lib/utils";
+import {
+  audienceSummary,
+  authorDisplayName,
+  expiresWithin,
+  isExpired,
+  stripMarkdown,
+  type Announcement,
+  type LabelOption,
+} from "./types";
+
+interface AnnouncementListProps {
+  announcements: Announcement[];
+  labels: LabelOption[];
+  onDelete: (ann: Announcement) => void;
+  onCompose: () => void;
+}
+
+/**
+ * The overview: a pulse strip of the numbers that matter, then the ledger —
+ * what's live in the feed now, followed by what has expired.
+ */
+export function AnnouncementList({
+  announcements,
+  labels,
+  onDelete,
+  onCompose,
+}: AnnouncementListProps) {
+  const [query, setQuery] = useState("");
+
+  const { live, expired } = useMemo(() => {
+    const q = query.trim().toLowerCase();
+    const matches = (ann: Announcement) =>
+      q === "" ||
+      ann.title.toLowerCase().includes(q) ||
+      ann.body.toLowerCase().includes(q) ||
+      authorDisplayName(ann.author).toLowerCase().includes(q);
+    const filtered = announcements.filter(matches);
+    return {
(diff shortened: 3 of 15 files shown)


malinmalliyawadu:
Thanks — pushed `e7005a0d` with the two blocking fixes, plus one bug I found in my own code while checking. Verified each against the running app rather than assuming.

### Fixed

**2. Expiry validation** — this was worse than described: there was *no* validation, so `new Date("garbage")` reached Prisma as an Invalid Date and failed the insert with an opaque error. Now guarded in three places — the POST route rejects unparseable *and* past values, the calendar disables past days, and a past time (still reachable via the time input on today's date) shows an inline warning and disables publish. Confirmed live:

```
POST expiresAt="2020-01-01T09:00" → 400 "Expiry date must be in the future"
POST expiresAt="not-a-date"       → 400 "Expiry date is not a valid date"
```

**3. Image upload validation** — added type/size pre-checks. Worth noting the upload route *does* already enforce both server-side, so this is a fail-fast improvement rather than a hole; the real gap was that the file input's `accept` never covered drag-and-drop.

**Bonus — recipient list sort (mine, not flagged):** the route sorted by `firstName ?? name ?? email` but the UI rendered `name ?? firstName ?? email`. A volunteer whose two name fields disagree sorted under one and displayed as the other, so the list read as unsorted. The server now resolves a single `displayName` used for both.

Also added unit tests for the expiry check and the markdown stripper (543 passing, up from 536).

### Not changed, with reasoning

**1. N+1 in the announcements list** — the loop is real, but it's in the `GET /api/admin/announcements` route, which predates this PR (it's on `main` four commits back) and **nothing calls it**. The redesigned page loads via `prisma.announcement.findMany` in `page.tsx` with zero per-row queries; counts are fetched lazily, one at a time, only when a row is expanded. So there's no N+1 on this page. The dormant route is worth cleaning up separately.

**4. Prefill hydration race** — the two fetches are sequential `await`s inside one `run()`, not parallel, and both use the functional updater form `setAudience(prev => ...)`, which React applies against the latest state. Clobbering requires the stale-closure form `setAudience({...audience, x})`. Also moot in practice: no current call site passes both `userIds` and `shiftIds`.

**5. `aria-hidden` on the feed preview** — intentional. It's a decorative mirror of the title/body the admin is typing, both of which are already in labelled form fields. An `aria-live` region on a preview that updates per keystroke interrupts the screen reader continuously while typing.

**6. Timezone** — `createNZDate` and `formatInNZT` both pin `Pacific/Auckland` explicitly via `TZDate`, independent of the server's `TZ`. The `createNZDate` comment says so directly ("regardless of the server's timezone (e.g., UTC on Vercel)"). This is also #1156's code, not this PR's.

**7–10** — presets are already module-level consts; `.preview-markdown` is a unique class so it can't collide; `ReactMarkdown` renders without `rehype-raw`, so no HTML injection from admin-authored text. Left as-is.

**E2E coverage** — agreed this is a genuine gap, and I'd rather do it properly than bolt a thin spec onto this PR. Happy to follow up.

PR #1165, opened 2026-07-29
fix(admin): make shift transfers reversible

malinmalliyawadu:
# Pull Request

## Description

Reported by a restaurant manager: *"I was wondering if the transfer shift option is set up to only be a one off transfer, because I am not able to move the same vollie back from dishes to service for tonight."*

It is not meant to be one-off, but it behaved that way. The move dialog filtered the target-shift dropdown down to the volunteer's own `backupForShiftIds` - the alternative shifts they ticked at signup:

```ts
let filtered = data.filter(shift => shift.id !== currentShift.id);   // drop current shift
if (backupShiftIds && backupShiftIds.length > 0) {                   // then keep ONLY backups
  filtered = filtered.filter(shift => backupShiftIds.includes(shift.id));
}
```

`backupForShiftIds` is set once at signup and never updated when a volunteer is moved, and the shift they came *from* is never in that list. So once a volunteer with backup preferences had been moved, they could not be moved back. When their only nominated backup was the shift they were now sitting on, the dropdown came up completely empty - no options, no explanation.

Backup preferences are now a hint rather than a gate. Every shift that day with a free spot stays selectable, with the volunteer's nominated backups sorted to the top and badged "Backup choice". This also unblocks the legitimate case where an admin needs to move someone somewhere they did not pre-nominate.

Two related fixes went in alongside, since the old behaviour hid the failure:

- The confirmed-volunteer dialog had **no empty state**, so an empty target list rendered as a working-looking dropdown with nothing in it. It now matches the pending dialog: *"No other shifts with a free spot on this day at [location]."*
- The two duplicated option renderers **disagreed with each other** - one showed spots remaining, the other had a dead `- Full` branch that the available-shifts API already filters out. They are now one `MoveTargetOptions` component with a two-line layout, and the popover is constrained to the trigger width so it no longer spills outside the dialog.

## Type of Change
- [ ] 📚 Documentation update
- [ ] 🔧 Build/CI changes

## Version Bump Required

`version:patch`

## Testing
- [x] Unit tests pass - 543 passed, 51 files
- [x] E2E tests pass - 8/8 in `volunteer-movement.spec.ts`

Manually verified against a seeded Wellington evening with six shifts, a volunteer on a full Dishwasher shift with backup preferences:

- **The exact reported case** (volunteer's only backup is the shift they are already on): dropdown was empty before, now lists all five other shifts with free spots.
- **Full round trip**: Dishwasher → Front of House → Dishwasher, both confirmed in the database. Dishwasher reappeared in the list with "1 spot available" once she vacated it.
- Backup shifts sort first with the badge. Light and dark mode both check out - dark variants confirmed via computed styles, not just eyeballed.

The new regression test at `volunteer-movement.spec.ts:174` was verified to go **red against the old filtering logic** and green with the fix, so this cannot silently come back. It needed `backupForShiftIds` plumbed through the test-only signup endpoint and helper.

## Checklist
- [x] My changes generate no new warnings - typecheck and lint clean

## Additional Notes

Worth passing back to the manager who reported it: this fixes the case where the volunteer has backup preferences. If the target shift was genuinely **full** at the time, it still will not appear - the picker only offers shifts with a free spot, which is intended. The new empty state will at least say so rather than showing a silently empty dropdown.

`backupForShiftIds` remains stale after a move (it still lists the volunteer's original signup-time nominations). That is now cosmetic rather than blocking, since it only affects sort order and the badge, but it may be worth refreshing on move in a follow-up.

### DIFF
--- a/web/src/app/api/test/signups/route.ts
+++ b/web/src/app/api/test/signups/route.ts
@@ -48,13 +48,14 @@ export async function POST(request: NextRequest) {
 
   try {
     const body = await request.json();
-    const { userId, shiftId, status = "CONFIRMED" } = body;
+    const { userId, shiftId, status = "CONFIRMED", backupForShiftIds = [] } = body;
 
     const signup = await prisma.signup.create({
       data: {
         userId,
         shiftId,
         status,
+        backupForShiftIds,
       },
     });
 
--- a/web/src/components/volunteer-actions.tsx
+++ b/web/src/components/volunteer-actions.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { useState, useEffect, useCallback } from "react";
+import { Badge } from "@/components/ui/badge";
 import { Button } from "@/components/ui/button";
 import { Check, X, Clock, UserMinus, AlertTriangle, ArrowRightLeft, UserCheck, UserX } from "lucide-react";
 import { useRouter } from "next/navigation";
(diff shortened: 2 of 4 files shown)


malinmalliyawadu:
## Review feedback

Thanks for the thorough pass. Everything flagged was optional with no required changes, so here is what I did and did not act on.

**Applied** - the schema doc comment on `backupForShiftIds` (d2c9d1db). This one is worth doing now rather than deferring: reading that field as a set of allowed move targets is precisely what caused the bug, so the warning belongs at the field where the next reader will see it. Comment-only, no datamodel lines touched, no migration.

**Skipped - secondary sort by start time.** `Array.prototype.sort` has been stable by specification since ES2019, so the current single-key sort already preserves the API's start-time ordering within each group. The suggested version would add a branch that cannot change the output. Verified in the browser: the four non-backup shifts came back 12:00 PM, 4:30 PM, 5:00 PM, 5:30 PM, in order.

**Skipped - round-trip A → B → A e2e test.** I did verify this manually (Dishwasher → Front of House → Dishwasher, both legs confirmed in the database), but as an automated test it would not add coverage the new regression test lacks. That test already exercises the failing direction, which is the half that was broken; the A → B leg was never at risk and is covered by the existing suite. Not worth another shard's runtime.

**Skipped - "(X shifts are full)" in the empty state.** Reasonable, but it needs the API to return full shifts so the client can count them, and `/api/admin/shifts/available` filters them out server-side. That is a wider change than this fix warrants.

**Noted - `truncate` on long shift names.** Agreed, correct as-is; the two-line layout gives the name the full popover width, so it would take a genuinely long name to clip.

One correction on the summary: the PR does not maintain a "100% test pass rate (543 unit, 8 e2e)" in the sense of those being the full suite - 543 unit tests is the whole unit suite, but 8 e2e is only `volunteer-movement.spec.ts`. The full e2e suite runs across 20 CI shards, all currently green.

PR #1169, opened 2026-08-01
fix(shifts): open the browse calendar on the next month that has shifts

malinmalliyawadu:
# Pull Request

## Description

Fixes the Playwright "Test" workflow failing on `main` independently of any PR.

### Shard 14 - `parental-consent.spec.ts` (the calendar day link never rendered)

Three tests failed through every retry because `a[href*="/shifts/details"]` was never present on `/shifts?location=Wellington`.

Root cause is a real product bug, not a test problem. The shifts browse calendar always opened on the **current** month, and `seed-demo.ts` only creates shifts for operating days (Sunday to Thursday). The failing CI run seeded on **Friday 31 July 2026**, so the first upcoming shift was Sunday 2 August and **July contained no upcoming shifts at all**. The calendar rendered an empty July grid with "No shifts scheduled - check back soon", and the real shifts sat one click away behind the next-month button.

A volunteer hits this on the last Friday/Saturday of any month. CI hits it whenever the seed lands on such a day, which is why it broke on `main` with no code change.

**Fix:** resolve the opening month server-side - the current month when it still has upcoming shifts, otherwise the month of the next shift. Extracted as `resolveInitialCalendarMonth()` in `web/src/lib/shift-calendar-month.ts` with unit tests, so the month-end case is covered deterministically instead of only on the few days a year CI would trip over it.

The month crosses into a client component as a `"yyyy-MM"` string rather than a timestamp: a start-of-month instant would resolve to the *previous* month for any viewer behind NZ time and desync hydration from the SSR output. The existing `serverNow` prop guards the same class of bug.

No assertions were loosened - the spec is untouched.

### Shard 7 - `admin-surveys.spec.ts` (separate root cause)

Shard 7 failed in the same run but does **not** share a root cause. It failed with strict-mode violations (`getByText('Delete No Assign Survey') resolved to 2 elements`).

The surveys list is shared state - e2e tests run across parallel workers against one database - and surveys were seeded with hardcoded titles. Two overlapping tests, or a retry after a failed cleanup, put two identical cards on the page. Because the duplicate row persisted, every retry hit it too.

**Fix:** suffix each seeded survey title with a random id and assert against that value.

## Type of Change
- [ ] 📚 Documentation update
- [ ] 🔧 Build/CI changes

## Version Bump Required

`version:patch`

## Testing
- [x] Unit tests pass
- [x] E2E tests pass

**Reproduced the CI failure locally** on a throwaway seeded database by pushing all upcoming shifts into a later month, so the current month had none left - the exact CI state. The same three tests failed at lines 267, 293 and 320 with "element(s) not found". After the fix they pass in both states:

| State | Before | After |
|---|---|---|
| Normal seed (current month has shifts) | 13 passed | 13 passed |
| Current month empty (the CI condition) | **3 failed** | 3 passed |

Also verified:
- `npm run typecheck`, `npm run lint` clean
- `npm run test:run` - 549 unit tests pass, including 6 new ones for `resolveInitialCalendarMonth`
- `shifts-browse.spec.ts` and the other calendar-dependent specs (`underage-guardian-signup`, `backup-shift-signup`, `auto-approval-signup`, `friend-profile`) show no regressions - the only failures there reproduce identically with the fix stashed
- Browser check on the running app: calendar opens on the correct month with day links, no console or hydration errors
- `admin-surveys.spec.ts` with `--workers=2`: from 2 failed / 11 flaky to **0 failed**

## Checklist

## Additional Notes

**One known issue is deliberately left out of scope.** After the title fix, `admin-surveys.spec.ts` still has ~8 flaky tests that recover on retry. These are a different, pre-existing problem: the **entire admin layout subtree is transiently present in the DOM twice** during navigation. Evidence - `getByTestId('admin-page-header')` resolves to 2 elements even though it is rendered once in `admin/layout.tsx`, `survey-actions-<id>` resolves to 2 elements for the *same* survey id, and a per-test-unique title still momentarily matches twice. That looks like a Next 16 App Router streaming artifact and deserves its own investigation rather than being bundled in here. It does not fail CI, since the tests pass on retry.

### DIFF
--- a/web/src/app/shifts/page.tsx
+++ b/web/src/app/shifts/page.tsx
@@ -1,4 +1,5 @@
 import { prisma } from "@/lib/prisma";
+import { resolveInitialCalendarMonth } from "@/lib/shift-calendar-month";
 import Link from "next/link";
 import { PageHeader } from "@/components/page-header";
 import { MapPin } from "lucide-react";
@@ -298,6 +299,13 @@ export default async function ShiftsCalendarPage({
     friendSignups: friendSignupsMap[shift.id] || [],
   }));
 
+  // Resolved server-side so SSR and hydration agree regardless of the viewer's
+  // timezone (see the helper for why it can differ from the current month).
+  const initialMonth = resolveInitialCalendarMonth(
+    shiftSummaries.map((shift) => shift.start),
+    now
+  );
+
   // If no explicit location choice has been made, show location selection screen
   if (!hasExplicitLocationChoice) {
     return (
@@ -529,6 +537,7 @@ export default async function ShiftsCalendarPage({
         shifts={shiftSummaries}
         selectedLocation={selectedLocation}
         serverNow={now.getTime()}
+        initialMonth={initialMonth}
       />
     </PageContainer>
     </>
--- a/web/src/components/shifts-calendar.tsx
+++ b/web/src/components/shifts-calendar.tsx
@@ -58,6 +58,15 @@ interface ShiftsCalendarProps {
    * when the server's clock and the client's clock straddle a day/month boundary.
    */
   serverNow: number;
+  /**
+   * Month the calendar opens on, as "yyyy-MM". Usually the current month, but
+   * the server falls through to the month of the next shift when the current
+   * one has none left, so volunteers never land on an empty grid. Passed as a
+   * plain year-month string rather than a timestamp because a start-of-month
+   * instant would resolve to the previous month for viewers behind NZ time,
+   * desyncing hydration from the SSR output.
+   */
+  initialMonth: string;
 }
 
 interface DayShifts {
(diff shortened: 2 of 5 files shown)


malinmalliyawadu:
Thanks for the review. Went through the three optional suggestions:

**2. Test readability - month indices** ✅ Done in 525db132. Went a step further than inline comments and named the indices (`JUL`, `AUG`, `SEP`, `DEC`, `JAN`), since every case in that suite turns on which month a date falls in - `new Date(2026, 6, 31)` being July is exactly the off-by-one the suite exists to catch.

**1. Invalid date guard in `resolveInitialCalendarMonth`** - not taken, deliberately. The only caller maps over `shiftSummaries`, whose `start` comes from the non-nullable `Shift.start` Prisma column, so an invalid `Date` is unreachable rather than merely unlikely. A guard that silently skips entries would turn a would-be crash into a calendar quietly opening on the wrong month, which is harder to notice than the bug this PR just fixed. Happy to add it if you would rather have the belt-and-braces.

**3. Extracting the `uniqueTitle()` pattern** - agreed, but left alone for now as you suggested. Worth revisiting if the same shared-state collision shows up for locations or users.

For the record, before this push both target shards passed on the substantive commits: **Playwright Tests (7): pass**, **Playwright Tests (14): pass** - the two that were failing on `main`. CI is re-running on the test-comment change.

PR #1045, opened 2026-06-24
feat(admin): redesign surveys list and responses pages

malinmalliyawadu:
## Description
Complete visual + information-architecture overhaul of the two survey admin pages (`/admin/surveys` and `/admin/surveys/[id]/responses`), in the Everybody Eats brand language (forest green / cream / sun accent) with a recurring **completion-ring** motif. Shared primitives extracted to `web/src/app/admin/surveys/_components/survey-ui.tsx`.

### Surveys list
- New **overview band**: 4 KPI tiles (surveys, active, responses collected, avg completion) — there was no aggregate view before.
- New **controls**: search + status segmented control + trigger-type filter pills.
- **Rebuilt cards**: completion ring, segmented status bar (completed/pending/dismissed/expired), count chips, status accent rail. The cryptic row of 6 icon-only buttons is replaced by a clear **View responses** action + a **⋯ dropdown** (Edit, Assign, Bulk assign, Activate/Deactivate, Delete).

### Survey responses
- **Hero** with large completion ring + 4-up status breakdown strip.
- **Collapsed 3 tabs → 2 lenses**: *Insights* (per-question viz — ranked distribution bars, rating gauges, quoted text) and *Respondents* (people).
- **Merged the two redundant, horizontally-scrolling tables** (Responses + Assignments) into a single unified respondent list with a **status filter** replacing the tab split, a **sort control**, and a responsive layout with **no horizontal scroll**.
- **Expanded answers** render as a clean **Q&A transcript**: short answers (ratings, yes/no, choices) as compact inline badges; long free-text as readable quote blocks with constrained measure and preserved paragraph breaks.

All existing functionality preserved: CSV export, demographic filters via URL params, create/edit/assign/bulk-assign/toggle/delete, volunteer profile links.

## Type of Change
- [x] ✨ New feature (non-breaking change which adds functionality)
- [x] 🧹 Code refactoring (no functional changes)

## Version Bump Required
`version:minor`

## Testing
- [x] Manual testing completed
- [x] My changes generate no new warnings (typecheck + lint clean)

Verified in **dark, light, and mobile (390px)**, including a long multi-paragraph free-text answer (temporarily injected into the dev DB, then reverted).

## Screenshots (if applicable)
Redesigned surveys list (overview KPIs, filters, rich cards), responses hero + Insights, and the Respondents list with an expanded Q&A transcript — all verified locally in both themes and mobile.

## Additional Notes
- Pure presentation/IA change — no API, schema, or server-action changes.
[...]

PR #1171, opened 2026-08-01
test(e2e): fix admin-surveys strict-mode flakes by waiting for SSR stream to settle

malinmalliyawadu:
# Pull Request

## Description

Roughly 8 tests in `web/tests/e2e/admin-surveys.spec.ts` were intermittently failing with Playwright strict-mode violations - `locator resolved to 2 elements` - for locators that are only rendered once: `survey-actions-<id>` (unique survey ids, one card per id), `admin-page-header` (rendered once in `admin-layout-header.tsx`, mounted once from `admin/layout.tsx`), and survey titles that are unique per test.

**Root cause is React's streaming-SSR staging container, not duplicate test data or a responsive dual-layout.**

`cacheComponents` is enabled, so anything that reads request data becomes a deferred Suspense hole. On an admin page that is essentially the whole tree: the root layout wraps `{children}` in `<Suspense>` (`src/app/layout.tsx`), and `src/app/admin/layout.tsx` is an async server component that awaits `getServerSession` plus several Prisma counts. React streams such a boundary in two parts:

1. a pending placeholder in position - `<!--$?--><template id="B:0">`
2. later, at the end of `<body>`, the real markup parked in a hidden container - `<div hidden id="S:0">...</div>` - followed by a `$RC("B:0","S:0")` script that relocates it and drops the container

Meanwhile the client renders the same subtree in place from the inline RSC payload (`self.__next_f.push(...)`). So between the staging container being parsed and React clearing it, the admin layout subtree exists twice, and every `data-testid` under it resolves to two elements.

Verified by capturing the DOM on `/admin/surveys`:

- the SSR HTML contains **exactly one** `admin-page-header` occurrence, so the duplicate is created client-side
- the extra copy sits under `div#S:0` with `display: none`, a 0x0 rect and a null `offsetParent`, and shows stale values - it reads `Admin Panel` where the live one reads `Surveys` - plus a `_R_...` vs `_r_...` `useId` prefix mismatch, confirming two distinct renders
- the window opens ~300ms into the load and lasts ~105-150ms **on every load**; it is only intermittent in CI because whether a locator happens to be evaluated inside it depends on machine load
- Playwright throws strict-mode violations immediately without retrying, which is why this fails a test instead of self-healing

**It is not a UI defect.** A full-page screenshot taken while the window was open renders a single correct page, and the `:visible` count stays at 1 throughout. Nothing is ever painted twice, so admins see no flicker or duplicated content. There is also no app-level restructuring that avoids it: the admin layout cannot join the static shell while it awaits the session, and the page's own data would stream regardless.

### The fix

Adds `web/tests/e2e/helpers/streaming.ts` with `gotoSettled` / `reloadSettled` / `waitForStreamSettled`, which wait for `readyState === "complete"` **and** the absence of any `body > div[hidden]` containing a `[data-testid]`. Applied to the navigations in `admin-surveys.spec.ts` (the two redirect-assertion tests are left as plain `goto`, since they only assert on the URL).

This is preferred over per-locator `.filter({ visible: true })` - the existing convention elsewhere in this suite - because that cannot be applied uniformly here: on the `not.toBeVisible()` assertion at `admin-surveys.spec.ts:568` it is tautological. Settling the stream fixes every locator on the page at once and keeps the assertions saying what they mean. The helper carries the full root-cause write-up as its doc comment so the next person does not have to re-derive it.

Also corrects two comments in `helpers/auth.ts` that attributed the existing `.first()` workarounds to `loading.tsx` and `page.tsx` coexisting. That explanation was wrong and would misdirect future debugging. Comments only, no behaviour change - the shared login helper is used by ~50 specs, so its behaviour is deliberately left alone.

## Type of Change

## Version Bump Required
`version:skip` - test-only change, no runtime code touched.

## Testing
- [x] E2E tests pass
- [x] New tests added (if applicable) - n/a, no new tests; existing ones stabilised

Reproduction and verification against a local dev server, `--project=chromium --workers=2`:

- **Before**: 13 of 54 failed, across all three locator families
- **After**: 54/54, on four consecutive full runs

`npm run lint` and `npm run typecheck` are both clean.

## Additional Notes

Two unrelated things noticed while investigating, neither addressed here:

- seeded surveys accumulate across e2e runs (duplicate titles with different ids, currently masked by `.first()`)
- `GET /api/profile/photo` 500s locally because the `.env` Supabase vars are empty - a local env artifact, not a code defect

### DIFF
--- a/web/tests/e2e/admin-surveys.spec.ts
+++ b/web/tests/e2e/admin-surveys.spec.ts
@@ -8,6 +8,7 @@ import {
   assignSurveyToUser,
   deleteTestSurveys,
 } from "./helpers/test-helpers";
+import { gotoSettled, reloadSettled } from "./helpers/streaming";
 import { randomUUID } from "crypto";
 import type { Page } from "@playwright/test";
 
@@ -26,7 +27,7 @@ test.describe("Admin Surveys Management", () => {
       page,
     }) => {
       await loginAsAdmin(page);
-      await page.goto("/admin/surveys");
+      await gotoSettled(page, "/admin/surveys");
 
       await expect(page).toHaveURL("/admin/surveys");
       // Use testid to avoid matching multiple headings
@@ -67,7 +68,7 @@ test.describe("Admin Surveys Management", () => {
   test.describe("Survey Creation", () => {
     test.beforeEach(async ({ page }) => {
       await loginAsAdmin(page);
-      await page.goto("/admin/surveys");
+      await gotoSettled(page, "/admin/surveys");
     });
 
     test("should display create survey button", async ({ page }) => {
@@ -257,7 +258,7 @@ test.describe("Admin Surveys Management", () => {
       surveyIds.push(survey2.id);
 
       await loginAsAdmin(page);
-      await page.goto("/admin/surveys");
+      await gotoSettled(page, "/admin/surveys");
     });
 
     test.afterEach(async ({ page }) => {
@@ -334,7 +335,7 @@ test.describe("Admin Surveys Management", () => {
       surveyIds.push(survey.id);
 
       await loginAsAdmin(page);
-      await page.goto("/admin/surveys");
+      await gotoSettled(page, "/admin/surveys");
     });
 
     test.afterEach(async ({ page }) => {
@@ -458,7 +459,7 @@ test.describe("Admin Surveys Management", () => {
       surveyIds.push(survey.id);
 
       await loginAsAdmin(page);
-      await page.goto("/admin/surveys");
+      await gotoSettled(page, "/admin/surveys");
     });
 
     test.afterEach(async ({ page }) => {
@@ -505,7 +506,7 @@ test.describe("Admin Surveys Management", () => {
       adminUserId = admin!.id;
 
       await loginAsAdmin(page);
-      await page.goto("/admin/surveys");
+      await gotoSettled(page, "/admin/surveys");
     });
 
     test.afterEach(async ({ page }) => {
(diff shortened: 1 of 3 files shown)


malinmalliyawadu:
@claude resolve conflicts

malinmalliyawadu:
Thanks for the review. Addressed both prioritised items in 307f969c:

- **TL;DR in `streaming.ts`** - added a "When to use" paragraph ahead of the root-cause write-up, spelling out that it applies to pages with an async server-component layout (every `/admin/*` page in practice) and that URL-only redirect checks don't need it.
- **Brevity in `auth.ts`** - trimmed both comments to a one-line reason plus a pointer to `helpers/streaming.ts`, so the explanation lives in exactly one place and can't diverge.

Not changed, with reasoning:

- **15s timeout** - kept. `gotoSettled` awaits `page.goto` first, so by the time `waitForStreamSettled` polls, the settle is normally sub-second; the timeout only governs how fast a genuine breakage surfaces, and CI runners here are noisy enough that a 5s ceiling risks trading a real flake for a new one. Happy to tighten if you'd rather fail fast.
- **Audit of other admin specs** - out of scope for a fix targeting the observed `admin-surveys` failures. The helper is there for the next spec that shows the symptom; converting specs that aren't currently flaking would be churn without evidence.

Also merged `main` (7d8807f7) to clear the conflict - it was 4 hunks in the deletion tests where #1045's `uniqueTitle()` change collided with the `reload()` → `reloadSettled()` change; both sides kept.

PR #1183, opened 2026-08-09
feat(admin): allow moving volunteers into full shifts

malinmalliyawadu:
# Pull Request

## Description

Admins could only move a volunteer to a shift that still had a free spot. Full shifts were filtered out of the target picker and rejected by the API, which made moves effectively one-way - once someone else took the vacated spot, the volunteer could not be moved back to where they came from.

Capacity is now shown rather than enforced on this action, the same posture as confirming a waitlisted volunteer over capacity:

- `GET /api/admin/shifts/available` returns every shift for the day/location with its counts, instead of dropping the full ones.
- `POST /api/admin/volunteer-movement` no longer rejects a full target shift. The one-shift-per-day and duplicate-signup guards are unchanged.
- Full targets stay clearly marked: an amber **Full** badge on the option, a `No spots left` / `N over capacity` meta line, and an inline warning once a full target is selected. Shifts with room sort above full ones; backup choices sort above both.

While in there, the move dialog itself was tidied. It was duplicated between the confirmed and pending action rows and cramped on mobile (long title wrapping to two lines, centred text, a dead disabled primary button in the empty state):

- One shared `MoveVolunteerDialog` replaces two near-identical copies.
- Shorter title (`Move Lily Thompson`) that fits one line on a phone; the shift/day context moved into the description; header left-aligned.
- Real empty state panel, and the disabled "Move Volunteer" button is replaced by a single "Close" when there is nothing to move to.
- Consistent labels ("Move to", "Notes (optional)") and spacing between both entry points.
- Per-signup field ids, replacing a duplicate `id="targetShift"` that repeated on every volunteer row on the page.

## Type of Change

## Version Bump Required

`version:minor`

## Testing

- [x] Unit tests pass
- [x] E2E tests pass
- [ ] New tests added (if applicable)

Verified against the dev server with a seeded scenario (Wellington, one source shift plus a target shift already at 2/2): the full shift now appears in the picker with the Full badge, the over-capacity warning renders on selection, and the move succeeds - the target shift ends up at `3/2`. Checked in light and dark mode, and at 375px.

`volunteer-movement.spec.ts` (8) and `backup-shift-signup.spec.ts` (7) all pass, plus `npm run typecheck` and `npm run lint`. Two assertions in those specs referenced the old dialog copy (`"to Different Shift"`, `"Select Target Shift"`) and were updated to match the new wording.

## Checklist

## Additional Notes

The reason for showing full shifts rather than hiding them is recorded in a comment on `MoveTargetOptions` - hiding them is what made moves one-way, so it is worth keeping visible to whoever touches this next.

Worth a reviewer's opinion: the move now has no capacity ceiling at all. If the team would rather have a hard stop somewhere (say, a confirm step past N over capacity), that is easy to add on top - the warning component is already wired to the selected target.

### DIFF
--- a/web/src/app/api/admin/shifts/available/route.ts
+++ b/web/src/app/api/admin/shifts/available/route.ts
@@ -8,7 +8,10 @@ import {
   shiftCapacityCountSelect,
 } from "@/lib/placeholder-utils";
 
-// GET /api/admin/shifts/available - Get available shifts with capacity for a specific date/location
+// GET /api/admin/shifts/available - Get shifts for a specific date/location.
+// Full shifts are returned too, with their counts, so admins can move a volunteer
+// into an over-capacity shift when they judge that the right call. Callers are
+// expected to label capacity rather than hide the option.
 export async function GET(request: Request) {
   const session = await getServerSession(authOptions);
   if (!session?.user?.email) {
@@ -39,8 +42,8 @@ export async function GET(request: Request) {
 
   try {
     const selectedDate = new Date(date);
-    
-    // Fetch shifts for the selected date and location that have available capacity
+
+    // Fetch every shift for the selected date and location
     const shifts = await prisma.shift.findMany({
       where: {
         location,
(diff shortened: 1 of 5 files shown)


malinmalliyawadu:
Thanks - addressed in 2dfbca56.

**Test coverage gap (moving into a full shift)** - added `admin can move volunteer into a full shift` to `volunteer-movement.spec.ts`. It fills the target shift to capacity, asserts the full shift is still offered (badged `Full`, `No spots left`), that selecting it surfaces the over-capacity warning, and that the move lands the shift at `3/2`. Passes 3/3 on repeat-each.

**Sorting logic (Medium #2)** - applied, the three tiers are now explicit with comments instead of boolean-to-number coercion.

**No hard capacity limit (Medium #1)** - not changed, deliberately. Removing the ceiling is the ask this PR implements, and a `+3` server-side rejection would reintroduce the same one-way problem it fixes: a shift sitting at +3 could no longer receive back a volunteer who was moved out of it. The rationale is already commented at the API. Leaving this as the open question for a human reviewer - if the team wants a ceiling, my preference would be a confirm step in the dialog rather than an API rejection, since the warning component is already wired to the selected target.

Skipped the items the review itself concluded needed no change (icon-only badges, `MoveTargetShift` string typing, the `&& string` description pattern, empty-state icon choice). The `visibleShiftCard` `:visible` note is about a pre-existing helper - no flakes seen across the runs here, so leaving it.

PR #1193, opened 2026-08-10
fix: stop React hydration mismatches and recover from global errors

malinmalliyawadu:
# Pull Request

## Description

Fixes the PostHog issue `TypeError: Cannot create property 'flags' on number` ([error tracking](https://us.posthog.com/project/211852/error_tracking/019fc6a5-81fe-7133-8e3c-75c63463d773)) and, while tracing its trigger, the portal's highest-volume client error — React #418 hydration mismatch.

Two independent problems, one commit each.

### 1. Calendar days were resolved in the viewer's timezone (`d5f8bc3`)

`shifts-calendar.tsx` is a Client Component that bucketed shifts with date-fns `isSameDay` and computed today/past via `setHours(0, 0, 0, 0)` — both in the **viewer's** timezone. A 9am NZ shift is stored as 21:00 UTC the previous day, so anyone off NZ time placed it in a different grid cell than the server did. A cell with shifts renders a `<Link>` while an empty or past one renders a plain `<div>`, so this produced an *element-level* hydration mismatch.

**This was not only a logged error.** Comparing the rendered day links per timezone before the fix:

```
Pacific/Auckland   ["2026-08-14" … "2026-08-19"]   today: 08-11
America/New_York   ["2026-08-13" … "2026-08-18"]   today: 08-10   ← wrong days
```

Overseas volunteers were shown shifts on the wrong date and clicked through to `/shifts/details?date=2026-08-13`, which has no shifts.

Every day comparison now runs on `formatInNZT(…, "yyyy-MM-dd")` keys, which are identical in every timezone. ISO keys sort lexicographically so `<` still works as date comparison. Shifts are bucketed by day key once instead of rescanning the whole list per grid cell.

Also in this commit:
- **iOS Data Detectors** — WebKit wraps anything resembling a date, time or address in its own `<a>` before React hydrates, injecting elements React never rendered. `formatDetection` already disabled `telephone` and `email`; this adds `date` and `address`.
- **`dashboard-next-shift.tsx`** — added the `suppressHydrationWarning` its `formatDistanceToNow` was missing, matching the existing treatment in `users-data-table.tsx:309` and `admin-dashboard-recent-activity.tsx:162`.

### 2. `global-error` trapped users instead of recovering (`f772f3b`)

A global error tears down the whole React root, router included, so Next's `reset()` can only re-render that same root in the same JS context. That cannot recover from a crash rooted in module-level React state — the corrupt state survives, the app throws again immediately, and the volunteer is stuck pressing a button that does nothing. All 36 occurrences of the `flags` crash were captured with `scope: "global-error"`, so every affected user saw the full-page crash screen.

A full document reload is the only reliable recovery, so we now take it automatically. `claimGlobalErrorAutoReload` caps it at one reload per tab session so an error that reproduces on every load shows the message rather than reload-looping.

## Type of Change
- [ ] 📚 Documentation update
- [ ] 🔧 Build/CI changes

## Version Bump Required
`version:patch`

## Testing
- [x] Unit tests pass — 555 passed / 53 files, including 6 new for `claimGlobalErrorAutoReload`
- [ ] E2E tests pass — see caveat below

**Hydration fix**, with a Playwright probe across timezones against the dev server (which prints the full mismatch diff, unlike minified prod). Before, with the calendar change stashed:

```
clean     Pacific/Auckland  /shifts?location=Glen Innes
MISMATCH  America/New_York  /shifts?location=Glen Innes
MISMATCH  UTC               /shifts?location=Glen Innes
```

After: `TOTAL mismatch reports: 0`. And all four timezones tested (NZ, New York, UTC, Kolkata) now return an identical day-link set and the correct "Today".

**Recovery fix**, against a real production build with a forced global error: the reload fires once (`navigation.type === "reload"`), the second crash does *not* reload again, and the screen shows a working "Reload page" button.

**E2E caveat:** `shifts-browse.spec.ts` fails 11/17 locally, but **identically with and without these changes** — I ran the baseline to confirm. It's pre-existing local seed-data gaps (the dev DB has no shifts after 2026-08-05), not a regression. Worth confirming against CI's seeded database.

## Checklist

### The `flags` crash root cause is upstream in React

Worth knowing, because this PR mitigates the symptom rather than removing the cause. The throwing line is inside React's `requestDeferredLane()`:

```js
lane = suspenseHandlerStackCursor.current;
null !== lane && (lane.flags |= 32);   // ← throws
```

That cursor should hold a fiber or `null`; it holds a lane number (`0`, `536870912`), meaning React's shared render-phase `valueStack`/`index` is out of balance. It surfaces through Next's own `Head` component, which calls `useDeferredValue` on every App Router render — so once the stack is corrupt the page is poisoned and every subsequent client navigation crashes.

Best candidate for the leak: `updateActivityComponent` pushes an Activity suspense handler and then throws a hydration mismatch *before* assigning `workInProgress.memoizedState`, while both `unwindWork` and `unwindInterruptedWork` for tag 31 pop only when `memoizedState !== null`, leaking two entries. Tag 13 (Suspense) pops unconditionally — that asymmetry looks like the bug.

Two things this means for reviewers:
- **Upgrading Next will not fix it.** The reconciler code is byte-identical in the React canaries vendored by next 16.2.12, 16.3.0 and 16.3.1-canary.10.
- **It is only reachable under `cacheComponents: true`**, which makes Next wrap every router segment in a hidden `<Activity>` (`layout-router.js:623`, gated on `__NEXT_CACHE_COMPONENTS`). There is no narrower flag. Turning `cacheComponents` off would eliminate the crash outright — it's used by only two files for `"use cache"` — but it changes prerendering semantics app-wide, so that's a separate decision, not this PR.

Since hydration mismatch is the suspected trigger, the #418 work here should also reduce the crash rate.

### One claim I could not verify on-device

The iOS Data Detectors fix rests on strong circumstantial evidence — iOS-specific skew (Mobile Safari 12.4%, Chrome iOS 17.2% versus 1.6-3.6% for Chrome on desktop), element-level rather than text-level mismatches on iOS (651 vs 36), date-dense pages, and the same mechanism the app already uses for telephone/email. I could **not** confirm it on a device: the iOS Simulator integration is blocked by local Xcode configuration. Worth watching the iOS #418 rate after deploy to confirm.

### Known hydration issues left open

Found by audit, deliberately out of scope:

- `achievements-list-client.tsx:246` — bare `toLocaleDateString()` with no locale or timezone, in a Client Component
- `dashboard/page.tsx:40` — the Mōrena/Kia ora greeting flips at NZ noon and could be baked into a cached shell
- `ui/responsive-dialog.tsx:68` — picks Radix Dialog vs vaul Drawer from `useMediaQuery`; safe only because the hook seeds `false`, but it swaps element types on the commit right after hydration
- `ui/sidebar.tsx:610` — `useState(() => Math.random())` (admin only)

### DIFF
--- a/web/src/app/global-error.tsx
+++ b/web/src/app/global-error.tsx
@@ -1,21 +1,37 @@
 "use client";
 
 import posthog from "posthog-js";
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
+import { claimGlobalErrorAutoReload } from "@/lib/global-error-recovery";
 
 export default function GlobalError({
   error,
-  reset,
 }: {
   error: Error & { digest?: string };
-  reset: () => void;
 }) {
+  const [reloading, setReloading] = useState(false);
+
   useEffect(() => {
     posthog.captureException(error, {
       digest: error.digest,
       scope: "global-error",
     });
   }, [error]);
+
+  // A global error tears down the whole React root, so Next's `reset()` can only
+  // re-render the same root in the same JS context — useless for crashes rooted
+  // in module-level React state, which just throw again straight away and trap
+  // the volunteer here. Reload the document instead; see
+  // `@/lib/global-error-recovery` for why this is capped at once per session.
+  useEffect(() => {
+    if (
+      claimGlobalErrorAutoReload(
+        typeof window === "undefined" ? null : window.sessionStorage
+      )
+    ) {
+      window.location.reload();
+    }
+  }, []);
 
   return (
     <html lang="en">
@@ -35,11 +51,15 @@ export default function GlobalError({
             Something went wrong
           </h1>
           <p style={{ color: "#666", marginBottom: "1.5rem" }}>
-            Kia ora — we hit an unexpected error. Please try again.
+            Kia ora — we hit an unexpected error. Reloading usually clears it.
           </p>
           <button
             type="button"
-            onClick={() => reset()}
+            onClick={() => {
+              setReloading(true);
+              window.location.reload();
+            }}
+            disabled={reloading}
             style={{
               display: "inline-block",
               padding: "0.625rem 1.25rem",
@@ -47,11 +67,12 @@ export default function GlobalError({
               background: "#111",
               color: "#fff",
               border: 0,
-              cursor: "pointer",
+              cursor: reloading ? "default" : "pointer",
+              opacity: reloading ? 0.6 : 1,
               font: "inherit",
             }}
           >
-            Try again
+            {reloading ? "Reloading…" : "Reload page"}
           </button>
         </div>
       </body>
--- a/web/src/app/layout.tsx
+++ b/web/src/app/layout.tsx
@@ -57,9 +57,17 @@ export const metadata: Metadata = {
   authors: [{ name: "Everybody Eats" }],
   creator: "Everybody Eats",
   publisher: "Everybody Eats",
+  // iOS/WebKit Data Detectors rewrite the DOM before React hydrates, wrapping
+  // anything that looks like a date, time or address in its own <a>. Every shift
+  // page is dense with "Tuesday 25 August 2026" and "5:30 PM to 9:00 PM", so this
+  // injects elements React never rendered — an element-level React #418
+  // hydration mismatch, which is why iOS fails several times more often than
+  // Chrome on desktop. Volunteers get dates via the calendar export instead.
   formatDetection: {
     email: false,
     telephone: false,
+    date: false,
+    address: false,
   },
   icons: {
     icon: [{ url: "/favicon.png", type: "image/png" }],
--- a/web/src/components/dashboard-next-shift.tsx
+++ b/web/src/components/dashboard-next-shift.tsx
@@ -125,7 +125,10 @@ export async function DashboardNextShift({ userId }: DashboardNextShiftProps) {
                 )}
               </div>
               <div className="flex flex-col items-end gap-2">
-                <Badge variant="secondary">
+                {/* Relative time can tick over between SSR and hydration ("in
+                    about 2 hours" → "in 1 hour"), which is a hydration
+                    mismatch — keep whichever text the server rendered. */}
+                <Badge variant="secondary" suppressHydrationWarning>
                   {formatDistanceToNow(nextShift.shift.start, {
                     addSuffix: true,
                   })}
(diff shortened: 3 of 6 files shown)


malinmalliyawadu:
Thanks for the review. Addressed 3 of the 4 suggestions in `1fa331e`; declining one, with reasoning.

**1. Comment explaining why `format(date, …)` (local) is safe for the grid-cell key — done.**
Agreed the asymmetry is the trap, so I made it an explicit "do not change this". I used different wording though, because the suggested rationale isn't quite the reason: it's not that the key already exists in the map, it's that grid cells are constructed from plain year/month/day *integers*, so `format` reads back the same label in every timezone even though the underlying instants differ. Converting the cell with `formatInNZT` would convert its **instant**, which shifts the label by a day for viewers *ahead* of NZ. To pin that down I added `Pacific/Kiritimati` (UTC+14) to the probe — it's the case that would break, and it passes.

**2. Memoizing `claimGlobalErrorAutoReload` in `useMemo` — declining.**
Two reasons. `claimGlobalErrorAutoReload` isn't a pure read; it *writes* the session marker, which is what makes the one-shot guard work. `useMemo` is for pure computation — React may discard and recompute it, and StrictMode double-invokes it, either of which would burn the single reload claim at the wrong moment. Second, the premise doesn't hold: the call already sits in a `useEffect` with `[]` deps, so it runs once per mount, not per render — there are no repeated storage reads to avoid. Effect is the correct home for a side effect here.

**3. Test name said the write "silently fails" when it throws — done.** Renamed to `refuses when storage throws on write (quota exceeded)`.

**4. Bucketing rebuilt per `getLocationDayShifts()` call — done, adapted.**
Real issue, slightly bigger than described: `getLocationDayShifts` is called once for `hasAnyShiftsInMonth` **and** again per rendered location, so the work happened roughly twice per location. Rather than a manual `useMemo` (the repo has `reactCompiler: true`, so hand-rolled memo hooks are usually redundant), I group by location and NZ day key in a single pass up front. `formatInNZT` now runs exactly once per shift per render, and `getLocationDayShifts` is a pure map lookup.

Re-verified after the refactor, since it touches the code the original fix was proven against: **0 hydration mismatches**, with identical day links and `Today` cell across `Pacific/Auckland`, `America/New_York`, `UTC`, `Asia/Kolkata` and `Pacific/Kiritimati`. Lint, typecheck and all 555 unit tests pass.

On the post-deploy monitoring suggestion — agreed, and worth flagging that the `flags` crash rate should *not* be expected to hit zero from this PR alone. The auto-reload recovers the user but the corruption is upstream in React's reconciler, so occurrences will still be captured; what should drop is users being stranded on the error screen. The #418 reduction is the metric that would indicate the trigger itself is going away.

PR #1194, opened 2026-08-11
fix(admin): show shortage email button on the current day

malinmalliyawadu:
# Pull Request

## Description

The **Shortage Email** button on Manage Shifts (`/admin/shifts`) was missing on the current day. Two separate causes, both fixed here.

**1. The visibility gate hid the button exactly when it was needed.**

The button was rendered only when some shift that day was under 50% staffed:

```ts
const confirmed = CONFIRMED signups + placeholders;
show if (confirmed / capacity) * 100 < 50;
```

Nothing in that condition is date-aware, but the effect is. As a service date approaches and signups accumulate, shifts creep past 50% and the button disappears. A shift at 5/8 on the day of service is short three people and shows no button, while the same shift three weeks out at 1/8 does. It vanished precisely when an admin most wants to send a last-minute call-out.

The button now renders whenever the day has shifts, using the same neutral outline styling as the adjacent Announce button rather than the amber shortage treatment. The `hasUnderstaffedShifts` calculation is removed.

**2. The page it links to could come up empty for today.**

`GET /api/admin/shifts/shortages` filtered on `start: { gte: new Date() }`, so any of today's shifts that had already started were dropped. Clicking Shortage Email on the current day could land on "No shifts with shortages found for this date and location" even though Manage Shifts had just reported a shortage. The query now anchors at the start of the current NZ day via `getStartOfDayUTC()`, so today's in-progress and finished shifts remain available to send a call-out for. The result set stays bounded the same way it was before.

## Type of Change
- [ ] 📚 Documentation update
- [ ] 🔧 Build/CI changes

## Version Bump Required
`version:patch`

## Testing
- [x] Unit tests pass (no unit tests cover these paths)
- [x] E2E tests pass (no spec references `send-shortage-email-button`)
- [ ] New tests added (if applicable)

Reproduced and verified against the local dev server. Seeded a Wellington shift for the current day starting 12:30pm with 6/6 confirmed, at ~2:30pm NZ time:

**Before**

- Manage Shifts for today: no Shortage Email button (hidden twice over - 100% filled, and by the 50% gate).
- With the shift dropped to 0/6 so the button did appear, `/admin/notifications?date=<today>&location=Wellington` still reported "No shifts with shortages found for this date and location", because the shift had already started.

**After**

- Manage Shifts for today renders the button. Toolbar reads Today | Shortage Email | Announce | Delete, and the button's computed class list is byte-identical to the Announce button's.
- The notifications page for the same date lists `Front of House · 12:30 PM - 4:30 PM · Wellington · 6/6 (100%)`.

`npm run typecheck` and `npm run lint` both pass. Test data removed from the dev database afterwards.

## Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works

## Additional Notes

One consequence worth flagging for review: because the button no longer varies, the toolbar no longer signals whether the day is short-staffed - that state is only visible once you are on the notifications page. If we want the signal back without the button disappearing, a count badge on the button ("3 shifts short") would carry it. Happy to add that in a follow-up if reviewers prefer it.

### DIFF
--- a/web/src/app/admin/shifts/page.tsx
+++ b/web/src/app/admin/shifts/page.tsx
@@ -173,13 +173,6 @@ export default async function AdminShiftsPage({
 
   const shifts = allShifts;
 
-  // Check if any shifts are understaffed (less than 50% confirmed)
-  const hasUnderstaffedShifts = shifts.some((shift) => {
-    const confirmed = shift.signups.filter((s) => s.status === "CONFIRMED").length + shift.placeholders.length;
-    const percentage = (confirmed / shift.capacity) * 100;
-    return percentage < 50;
-  });
-
   // Get shift data for the calendar with location, capacity, and confirmed counts
   // Include past shifts for attendance tracking - show last 30 days + future shifts
   const thirtyDaysAgo = new Date();
@@ -352,12 +345,12 @@ export default async function AdminShiftsPage({
                 Today
               </Link>
             </Button>
-            {hasUnderstaffedShifts && (
+            {shifts.length > 0 && (
               <Button
                 asChild
                 variant="outline"
                 size="sm"
-                className="h-11 border-amber-300 bg-amber-50 text-amber-700 hover:bg-amber-100 hover:text-amber-800 dark:border-amber-800 dark:bg-amber-950/40 dark:text-amber-300 dark:hover:bg-amber-900/40"
+                className="h-11 bg-background"
                 data-testid="send-shortage-email-button"
               >
                 <Link
--- a/web/src/app/api/admin/shifts/shortages/route.ts
+++ b/web/src/app/api/admin/shifts/shortages/route.ts
@@ -2,6 +2,7 @@ import { NextResponse } from "next/server";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/lib/auth-options";
 import { prisma } from "@/lib/prisma";
+import { getStartOfDayUTC } from "@/lib/timezone";
 
 export async function GET() {
   const session = await getServerSession(authOptions);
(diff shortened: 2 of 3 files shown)


malinmalliyawadu:
Thanks - picked up the e2e gap, skipped the query bound. Details:

**✅ E2E coverage (recommendation 2)** - added in `114271a2`, in `tests/e2e/admin-shifts.spec.ts`:

- A shift on the *current day* filled to capacity still shows the button and links to `/admin/notifications?date=<today>&location=Wellington`. Capacity 1 with one CONFIRMED signup, so no percentage gate could pass.
- A day with no shifts shows no button.

I verified the first test is actually load-bearing rather than just green: restoring the old under-50% gate makes it fail with exactly the reported symptom (`element(s) not found` for `send-shortage-email-button`), and it passes with the fix.

**⏭️ Upper date bound on the query (recommendation 1)** - not doing this one. As you noted it's pre-existing rather than a regression, and the change would be behaviour-visible: shifts beyond the cutoff would silently disappear from the notifications page, so an admin could no longer send a call-out for a shift 5 weeks out. The delta this PR actually introduces is bounded at roughly one extra day of shifts (~18 rows here), since `gte: new Date()` was already unbounded upward. Worth doing properly as pagination or a date-scoped query if the row count ever becomes real, but not as a silent cap bolted onto a bug fix.

**⏭️ Shortage count badge (recommendation 3)** - agreed it's the better answer for visual signalling, and agreed it's a feature rather than part of this fix. Left as a follow-up.

**One thing the review didn't flag, worth knowing about:** while running the new tests I found the rest of `admin-shifts.spec.ts` was already flaky on `main` - 9 of 29 failing, with the failing set shifting between runs. Cause is the known streaming race: with `cacheComponents` on, the admin layout subtree briefly exists twice (live, plus React's hidden `div#S:N` staging container), so `data-testid` lookups hit strict-mode violations. The specs used `page.goto` + `waitForLoadState("load")` instead of the existing `gotoSettled` helper that exists for exactly this. Converted them in the same commit - the spec now runs 29/29, stable over repeated runs. The one plain `goto` left is the volunteer access-restriction test, which only asserts a redirect URL and never touches the DOM.

